PR #4, opened 2022-12-12
Site layout fix

galund:
This change goes along with an update to the mkdocs-govuk theme - see https://github.com/alphagov/mkdocs-govuk/commit/f85333c897a7fdc2705c637230d2316599aeff32

Really not sure how this ever worked but I swear it did once :/

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -3,7 +3,7 @@ repo_url: https://github.com/alphagov/di-identity-vocab
 theme:
     name: null
     custom_dir: 'mkdocs-govuk/govuk_theme'
-tech_docs:
+
     show_govuk_logo: true
     service_link: https://www.sign-in.service.gov.uk/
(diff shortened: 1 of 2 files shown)


mkdocs.yml
```diff
@@ -3,7 +3,7 @@ repo_url: https://github.com/alphagov/di-identity-vocab
 theme:
     name: null
     custom_dir: 'mkdocs-govuk/govuk_theme'
-tech_docs:
```
cmiddi, line 6: So I think you need this providing you are on the correct branch of mkdocs-govuk - match-techdocs-template, if you are on this branch the template files refer to config.tech_docs throughout and if you don't have it the formatting doesn't work.  Looks like it may have been a problem with pulling the submodules down at the right branch.

cmiddi (approved):

PR #7, opened 2022-12-13
Build, test, deploy workflows/pipelines to publish to GH Pages

galund:
A pipeline that appears to work well enough to deserve some more review. We won't set up DNS for the GH Pages site until we've had some content design eyes and fixed the layout.

We're deploying to https://alphagov.github.io/di-identity-vocab/

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -1,6 +1,7 @@
-name: 'Build, test, deploy'
+name: 'Build and test'
 
 on:
+  workflow_call:
   workflow_dispatch:
   push:
     branches:
@@ -62,5 +63,3 @@ jobs:
         uses: actions/upload-pages-artifact@v1.0.6
         with:
           path: site
-
-
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -10,8 +10,6 @@ on:
 
 permissions:
   contents: read
-  # Needed for the 'trilom/file-changes-action' action
-  pull-requests: read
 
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,36 @@
+name: Deploy MkDocs site to Pages
+
+on:
+  push:
+    branches: [$default-branch, deploy-to-gh-pages]
```
philandstuff: guessing you'll remove the second branch before merging?

galund: Yep, should do :-) - impossible to test without tho!

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: Build
+        run: scripts/build
+
+      - name: Run all tests
+        run: scripts/test
+
+      - name: Upload HTML
+        uses: actions/upload-pages-artifact@v1.0.6
+        with:
+          path: site
```
TobySaundersGDS, line 54: Does this path match the output of the build script? I could't tell easily.

galund: Ah, yes it does, the line that does it is this https://github.com/alphagov/di-identity-vocab/blob/87ceeab90765256c54e94a8b16e9c698da5fb7f1/scripts/build#L16 and the default is "site" per https://www.mkdocs.org/user-guide/configuration/#site_dir

TobySaundersGDS (approved):
nice

TobySaundersGDS (approved):

PR #8, opened 2022-12-13
Change description to be address specific and change schema to addressbase

cmiddi:


v1/linkml-schemas/address.yaml
```diff
@@ -16,7 +16,7 @@ default_range: string
 
 classes:
   PostalAddressClass:
-    class_uri: schema:PostalAddress
+    class_uri: addressbase
```
galund, line 19: This needs to be a class, not just a schema that contains a class? So I feel like it's wrong, but it's difficult to be sure as the URI isn't resolvable anyway and I can't see any docs for their schema :-) ... I'm not sure I get that this is useful if their schema isn't actually published?

v1/linkml-schemas/address.yaml
```diff
@@ -1,10 +1,10 @@
 id: https://vocab.account.gov.uk/linkml/address-schema
 name: address-schema
 description: >-
-  A real name for a user, composed of an ordered list of given names and family names (etc)
+  A representation of an address based on the addressbase schema, with optional validity dates
 prefixes:
   linkml: https://w3id.org/linkml/
-  schema: https://schema.org/
+  addressbase: http://namespaces.geoplace.co.uk/addressbase/2.1
```
cmiddi, line 7:
```suggestion
  geoplace: http://www.geoplace.co.uk/addressbase/schema/2.1/addressbase.xsd
```

v1/linkml-schemas/address.yaml
cmiddi, line 19:
```suggestion
    class_uri: geoplace:Address
```

cmiddi (commented):
There is a xsd file available that is downloadable and then there is an element called Address within there hopefully that makes a bit more sense?

v1/linkml-schemas/address.yaml
galund: hmm, now I can see the schema, I can see how it attempts to define the namespace URI, in which case I think what you had originally was correct!

I'm not completely convinced this is the right thing to do, because I have this theory that these could be addresses from anywhere in the world, and AddressBase is very adamant that it's a UK-only thing (see "CountryCodeType" etc).

It would be good to reference geoplace for the specific slots that map over directly, like the UPRN, though.

cmiddi:
replaced by later pull request.

PR #9, opened 2022-12-14
new address changes based on addressbase specification

cmiddi:
Definitions found further down this document https://www.ordnancesurvey.co.uk/documents/product-support/tech-spec/addressbase-technical-specification.pdf show that they are from the gml schema.  Also dropped the change for making postcode a required field.

### DIFF
--- a/v1/linkml-schemas/address.yaml
+++ b/v1/linkml-schemas/address.yaml
@@ -6,6 +6,7 @@ prefixes:
   linkml: https://w3id.org/linkml/
   schema: https://schema.org/
   di_vocab: https://vocab.account.gov.uk/v1/
+  gml: http://www.opengis.net/gml
 imports:
   - linkml:types
   - ./common
@@ -36,15 +37,28 @@ classes:
 
 slots:
   uprn:
+    slot_uri: gml:uprn
   organisationName:
+    slot_uri: gml:organisationName
   departmentName:
+    slot_uri: gml:departmentName
   subBuildingName:
+    slot_uri: gml:subBuildingName
   buildingNumber:
+    slot_uri: gml:buildingNumber
   buildingName:
+    slot_uri: gml:buildingName
   dependentStreetName:
+    slot_uri: gml:dependentThoroughfare
   streetName:
+    slot_uri: gml:thoroughfare
   doubleDependentAddressLocality:
+    slot_uri: gml:doubleDependentLocality
   dependentAddressLocality:
+    slot_uri: gml:dependentLocality
   addressLocality:
+    slot_uri: gml:postTown
   postalCode:
+    slot_uri: gml:postcode
   addressCountry:
+    slot_uri: schema:addressCountry
--- a/v1/linkml-schemas/address.yaml
+++ b/v1/linkml-schemas/address.yaml
@@ -6,7 +6,7 @@ prefixes:
   linkml: https://w3id.org/linkml/
   schema: https://schema.org/
   di_vocab: https://vocab.account.gov.uk/v1/
-  gml: http://www.opengis.net/gml
+  adb: http://namespaces.geoplace.co.uk/addressbase/2.1
 imports:
   - linkml:types
   - ./common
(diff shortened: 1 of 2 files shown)


v1/linkml-schemas/address.yaml
```diff
@@ -36,15 +37,28 @@ classes:
 
 slots:
   uprn:
+    slot_uri: gml:uprn
```
galund: I still can't see how these resolve to anything relative to the `gml` definition given above, I can buy that they actually don't, but if so then we should put a link into some documentation that defines the URIs.

cmiddi: Yeah had a further look after we discussed the other day.  I need to look again.

v1/linkml-schemas/address.yaml
```diff
@@ -6,6 +6,7 @@ prefixes:
   linkml: https://w3id.org/linkml/
   schema: https://schema.org/
   di_vocab: https://vocab.account.gov.uk/v1/
+  gml: http://www.opengis.net/gml
```
cmiddi:
```suggestion
  adb: http://namespaces.geoplace.co.uk/addressbase/2.1
```

v1/linkml-schemas/address.yaml
cmiddi:
```suggestion
    slot_uri: adb:uprn
```

cmiddi: added into new file commit

v1/linkml-schemas/address.yaml
cmiddi: added into new file commit

v1/linkml-schemas/address.yaml
```diff
@@ -36,15 +37,28 @@ classes:
 
 slots:
   uprn:
+    slot_uri: adb:uprn
```
cmiddi:
```suggestion
    range: integer
    slot_uri: adb:uprn
```

cmiddi: uprn defined as an integer by addressbase and we are passing around as an integer based on data gathered from integration environment.

galund: I have a feeling this would actually be a breaking change as the address CRI is probably putting UPRN into a string in its JSON, but we could check with the Orange team.

cmiddi: From the staging environment just now:

`    "credentialSubject": {
      "address": [
        {
          "addressCountry": "GB",
          "uprn": 100120012077,
          "buildingName": "",
          "streetName": "HADLEY ROAD",
          "postalCode": "BA2 5AA",
          "buildingNumber": "8",
          "addressLocality": "BATH",
          "validFrom": "2000-01-01",
          "subBuildingName": ""
        }
      ]
    }`

be good to check with the Orange team though.

v1/linkml-schemas/address.yaml
```diff
@@ -36,15 +37,29 @@ classes:
...
   streetName:
+    slot_uri: adb:thoroughfare
   doubleDependentAddressLocality:
+    slot_uri: adb:doubleDependentLocality
   dependentAddressLocality:
+    slot_uri: adb:dependentLocality
   addressLocality:
+    slot_uri: adb:postTown
   postalCode:
+    slot_uri: adb:postcode
```
cmiddi: Are we in a position while we accept only UK addresses to provide a pattern against the postalCode that we provide?  The addressbase schema has a regular expression defined currently that enforces a space in the middle of the post code.  This would all depend on whether we have similar validation in the user interface should the user decide to overtype a lookup or provide their postcode manually.

v1/linkml-schemas/address.yaml
```diff
@@ -36,15 +37,29 @@ classes:
...
+    range: integer
+    slot_uri: adb:uprn
   organisationName:
+    slot_uri: adb:organisationName
   departmentName:
+    slot_uri: adb:departmentName
   subBuildingName:
+    slot_uri: adb:subBuildingName
   buildingNumber:
+    slot_uri: adb:buildingNumber
```
cmiddi: buildingNumber is defined as an integer in the addressbase schema, but we seem to be using as a string, so I believe to change this would cause a breaking change.  We should maybe review the slot_uri property if we are not mapping directly to the same type, or we should provide an explanation that we are using the string representation of what is defined in addressbase.

v1/linkml-schemas/address.yaml
galund: On the grounds that some users may at some point be allowed to provide a non-UK address for certain purposes, I think it'd be better to avoid bringing in validation of post codes. That doesn't preclude us from linking to addressbase in terms of a slot URI - but in this case I think it might be more clear just to to refer to https://schema.org/postalCode

cmiddi:
```suggestion
    slot_uri: schema:postalCode
```

v1/linkml-schemas/address.yaml
galund: Yep, not sure the right thing to do, but we could always record it as an issue that could be resolved in a "version 2" one day. I think maybe slot_uri referencing ADB is fine without having the same exact schema, it's more about whether the semantics are the same - but either way a clarification note against the field would be a good idea.

v1/linkml-schemas/address.yaml
```diff
@@ -36,15 +37,29 @@ classes:
...
   dependentStreetName:
+    slot_uri: adb:dependentThoroughfare
   streetName:
+    slot_uri: adb:thoroughfare
   doubleDependentAddressLocality:
+    slot_uri: adb:doubleDependentLocality
   dependentAddressLocality:
+    slot_uri: adb:dependentLocality
   addressLocality:
+    slot_uri: adb:postTown
```
galund: I think I prefer to reference schema.org for anything where there's a direct mapping as addressbase definitions tend to mention Royal Mail, which feels weird for something that's supposed to be international.

However, this reminds me that we need to look into whether we should be using `close_mappings` and `exact_mappings` etc rather than `slot_uri` for some of these cases, and what the effect might be on our JSON-LD should we ever get that working.

Example of that is @ https://github.com/linkml/linkml/blob/main/examples/PersonSchema/personinfo.yaml

v1/linkml-schemas/address.yaml
cmiddi: new version pushed shows buildingNumber as a close_mapping and a note added in description that it is a string representation of the field.

v1/linkml-schemas/address.yaml
cmiddi: I agree slot_uri is probably not the right thing where we are not 100% the same thing.  I have pushed a new version to use exact_mappings and close_mappings where appropriate.

galund (approved):
This feels right, we may well need to review again once we've figured out how to make the JSON-LD work.

PR #10, opened 2022-12-20
Fix layout and update build to use SASS and webpack

galund:
- bring in JS, CSS and other assets from mkdocs-govuk and its dependency tech-docs-assets
- we're on a branch of mkdocs-govuk whilst we check this builds right
- ajv-formats is probably not required as a dependency if we're installing it globally


galund:
This does look like it's working (have done a manual deployment to <https://alphagov.github.io/di-identity-vocab/>) so I'm going to get the dependency chain sorted out (ie merged onto their respective main branches).

cmiddi (approved):
This branch works with the associated mkdocs-govuk updates from a fresh checkout.

PR #11, opened 2022-12-23
Proposed changes for the authorization request class

cmiddi:
Looking to bring some of the documentation elsewhere into the vocab site.  Also to include missing properties such as vtr, nonce and scope.

### DIFF
--- a/v1/examples/core-cri-authz-request.json
+++ b/v1/examples/core-cri-authz-request.json
@@ -2,7 +2,7 @@
   "aud": "http:\/\/www.example.com\/audience",
   "sub": "test-user-id",
   "nbf": 1659007190,
-  "shared_claims": {
+  "claims": {
     "address": [
       {
         "departmentName": "someDept",
@@ -92,5 +92,7 @@
   "state": "S-mX7O17YDxkCTGxxTg68nWW1Of6KhnkVrK1uZlG76s",
   "exp": 1659008090,
   "iat": 1659007190,
-  "client_id": "ipv-core"
+  "client_id": "ipv-core",
+  "scope": "openid",
+  "nonce": "1233123213"
 }
--- a/v1/examples/core-cri-authz-request.json
+++ b/v1/examples/core-cri-authz-request.json
@@ -2,7 +2,7 @@
   "aud": "http:\/\/www.example.com\/audience",
   "sub": "test-user-id",
   "nbf": 1659007190,
-  "claims": {
+  "shared_claims": {
     "address": [
       {
         "departmentName": "someDept",
--- a/v1/examples/rp-authz-request.json
+++ b/v1/examples/rp-authz-request.json
@@ -1,0 +1,10 @@
+{
+  "claims": "%7B%22userinfo%22%3A%7B%22https%3A%2F%2Fvocab.account.gov.uk%2Fv1%2FcoreIdentityJWT%22%3A%7B%22essential%22%3Atrue%7D%7D%7D",
+  "response_type": "code",
+  "redirect_uri": "http://localhost:3001/oauth/callback",
+  "state": "tRNlWtowYYcSBUzFfkjXbn8GX0V4RansaA2USbjZmlQ",
+  "client_id": "IcVM2g9s10dXL0Jpd4_rVfQ_G0Y",
+  "scope": "openid email phone offline_access",
+  "nonce": "1233123213",
+  "vtr": ["P2.Cl.Cm"]
+}
--- a/v1/linkml-schemas/credentials.yaml
+++ b/v1/linkml-schemas/credentials.yaml
@@ -27,6 +27,7 @@ classes:
       - exp
       - aud
       - iat
+      - jti
   AuthorizationRequestClass:
     is_a: JWTClass
     slots:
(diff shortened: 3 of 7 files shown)


v1/examples/core-cri-authz-request.json
```diff
@@ -2,7 +2,7 @@
   "aud": "http:\/\/www.example.com\/audience",
   "sub": "test-user-id",
   "nbf": 1659007190,
-  "shared_claims": {
+  "claims": {
```
galund: Nope, it's called `shared_claims` per https://github.com/alphagov/digital-identity-architecture/blob/main/rfc/0031-ipv-generic-frontend-cri-interface.md#requests-to-the-authorization-endpoint

(I didn't like this, but it's not the same format as an [OpenID Connect `claims`](https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter) so it's fine that it be called something different. I might have preferred to call it `credentialSubject` maybe.)

cmiddi: OK cool my misunderstanding, so we use `shared_claims` internally to pass data, but what we get from a Relying Party is an OpenID Connect `claims` value to let us know what they expect back from the `/userinfo` endpoint, so we need to support both as optional things?  Once added I could another test .json file for a sample external relying party request?

galund: Ah well this file is *CRI* authz request, so if we're documenting the RP interface that'd be somewhere else.

cmiddi: Yeah I will modify the underlying AuthorizationRequestClass, so it can support both `shared claims` and `claims` and introduce a new test file rp-authz-request.json and make that reflect what the RP needs to send to the authorize endpoint.  This should hopefully make the AuthorizationRequestClass generic enough that it can be used for both purposes.

cmiddi: Updates added to put shared_claims back in and added support for claims from a Relying Party call to the authorize endpoint

v1/linkml-schemas/credentials.yaml
```diff
@@ -31,16 +31,25 @@ classes:
...
-      - shared_claims
       - response_type
       - scope
       - client_id
       - state
       - redirect_uri
       - nonce
-      - prompt
+
+  RPAuthorizationRequestClass:
```
galund: We've named these two things inconsistently. Here, the name is for the client (ie the RP), but in the `CRIAuthorizationRequestClass` case, you've named it for the server (the CRI).

Perhaps we should look for consistency with OpenID Connect, for the RP case, perhaps call it [`OpenIDConnectAuthenticationRequest`](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).

In the CRI case, my starter for 10 would be `IssuerAuthorizationRequest` as I want to avoid "CRI" as an abbreviation wherever possible, because it causes confusion with "CI" for contra-indicator.

cmiddi: Class names updated based on comments.

v1/examples/rp-authz-request.json
```diff
@@ -0,0 +1,10 @@
+{
+  "claims": "%7B%22userinfo%22%3A%7B%22https%3A%2F%2Fvocab.account.gov.uk%2Fv1%2FcoreIdentityJWT%22%3A%7B%22essential%22%3Atrue%7D%7D%7D",
+  "response_type": "code",
+  "redirect_uri": "http://localhost:3001/oauth/callback",
+  "state": "tRNlWtowYYcSBUzFfkjXbn8GX0V4RansaA2USbjZmlQ",
+  "client_id": "IcVM2g9s10dXL0Jpd4_rVfQ_G0Y",
+  "scope": "openid email phone offline_access",
```
galund, line 7: I don't think we do `offline_access` but it might be worth pinging @mrwilson to see if there is a sample/test fixture RP Authz request we can/should be keeping this in sync with.

cmiddi: This example I think was taken from the RPStub that @zapthedingbat had put together and it is contained in some documentation that I have seen from @philf999, but I will check with @mrwilson.

mrwilson: We do support `offline_access`

https://github.com/alphagov/di-authentication-api/blob/main/shared/src/main/java/uk/gov/di/authentication/shared/entity/ValidScopes.java#L17

galund: Hmmmmm okay, does that work for identity proofing data? The VC from SPOT should expire pretty rapidly after going through so this seems somewhat dubious to me.

mrwilson: No, it doesn't - we can only pull the SPOT output (on a call to /userinfo) once, then we delete after first read. This is more for ID tokens than actual identity info.

galund: We should probably make this clear in the docs if it isn't already - for me that would be reason enough to decide *not* to support it, but maybe there are use cases I haven't thought about. Happy to resolve this convo here but I'll cc @zapthedingbat as this covers RP happiness :-)

galund: Also cc @philf999

philf999: Myself and @zapthedingbat  have wondered what the refresh token would be used for. As mentioned the claims that are retrieved from userinfo expire as quickly as the access_token. I'm not sure if there is any use case for refreshing an id_token.

mrwilson: There's ongoing discussion in the context of VC storage where one of the solutions involves access tokens issued from Orchestration to IPV Core (this is in no way the settled solution for this, fwiw)

galund: Okay cool, I'll resolve this conversation here, but if you don't have it captured in your backlog (ie "consider not supporting this, or explaining more clearly in the docs why RPs should probably not use it"), then please do feel free ;-)

v1/linkml-schemas/credentials.yaml
```diff
@@ -73,31 +86,68 @@ slots:
     range: PersonWithOnlyCoreAttributesClass
   iss:
     range: uri
+    description: The `iss` (issuer) claim identifies the principal that issued the JWT. The processing of this claim is generally application specific. [[RFC 7519](https://www.rfc-editor.org/rfc/rfc7519), [Section 4.1.1](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.1)]
```
galund: I'm not sure that the "application specific" thing is massively helpful for us to include here, but I guess there subtle shades between JWTs in OAuth vs JWTs that are VCs.

I do think we should capture something about validation of issuers against key material in *our* system, as that process is an important part of our security architecture (ie checking that `iss` is as expected given `kid`), especially in the context of verifiers that wrangle credentials from multiple issuers (ie SPOT and IPV Core). See <https://govukverify.atlassian.net/browse/SPT-131>.

Gonna play around with some wording here...

cmiddi: Did you manage to get anywhere with proposing some wording here? - would be good to try and get this merged if we can.

cmiddi: minimised the comments on the JWT claims so we can get this merged for now.

galund (approved):

PR #13, opened 2023-01-24
Document details updates

cmiddi:
Initial updates to bring the document.yaml file in line with current implementation.  Using personalNumber for driving licenses and documentNumber for passport.  Also added residencyPermit in line with RFC-0024 document updates, these are expected to be in line with the same model for passport.


v1/linkml-schemas/person.yaml
```diff
@@ -53,4 +54,8 @@ slots:
   drivingPermit:
     range: DrivingPermitDetailsClass
     multivalued: true
-    inlined_as_list: true
\ No newline at end of file
+    inlined_as_list: true
```
galund, line 58: We should check why we're doing this for drivingPermit but not for passport? I think they should be consistent.

cmiddi: Yeah I will take a look, I think it was probably just missing from the beginning, but no real reason why.  If I can't find any evidence as to why it needs to be different I will update.

v1/linkml-schemas/document.yaml
```diff
@@ -20,24 +20,50 @@ classes:
...
+    description: An identifier for the issuing authority, for example, DVLA.
   issueNumber:
-    range: string
\ No newline at end of file
+    description: An identifier that changes with each issue of the document - unique only within documents issued under the same `personalNumber`.
+  fullAddress:
+    description: Unparsed address string as retrieved from optical character recognition read of a document.
+  icaoIssuerCode:
+    description: An identifier for the issuing State or Organisation of a passport as defined by the [International Civil Aviation Organization (ICAO) Standard for Machine Readable Travel Documents](https://www.icao.int/publications/Documents/9303_p3_cons_en.pdf). Usually three characters.
+    required: true
```
galund, line 67: We should probably check when we actually added this field and whether there might be some VCs in the wild that don't actually have it.

galund (approved):

PR #14, opened 2023-01-30
draft changes for the name data model

cmiddi:
Initial attempt at some description and formatting changes in the way names are displayed.

### DIFF
--- a/v1/linkml-schemas/name.yaml
+++ b/v1/linkml-schemas/name.yaml
@@ -17,34 +17,37 @@ default_range: string
 
 classes:
   NameClass:
+    description: Name object that represnts a user's claimed identity.  This will have a list of name parts.  Usually this will contain one or more [GivenNameClass](../GivenNameClass) name parts and one [FamilyNameClass](../FamilyNameClass) name part for a specific period of validity.
     mixins:
       - ValidityClass
-    attributes:
-      nameParts:
-        range: NamePartClass
-        multivalued: true
-        inlined_as_list: true
+    slots:
+      - nameParts
+      
   NamePartClass:
+    description: Abstract class to define a name part value and name part type for an optional validity period.
     abstract: false  # has to be false otherwise JSON-Schema doesn't include it, and validation fails :(
     mixins:
       - ValidityClass
     slots:
       - value
-    attributes:
-      type:
-        range: NamePartType
-        multivalued: false
-  GivenNameClass:
-    is_a: NamePartClass
-    attributes:
-      type:
-        equals_string: GivenName
+      - type
 
-  FamilyNameClass:
+  GivenNameClass:
+    description: Name part object that represents a Given Name.  There may be multiple [NamePart](../NamePartClass) objects with a [NamePartType](../NamePartType) of `GivenName` within the representation of a [Name](../NameClass) for a specific period of validity.  These will appear in the order in which the user specified or the order in which they were read from official documentation.
     is_a: NamePartClass
-    attributes:
-      type:
-        equals_string: FamilyName
+
+  FamilyNameClass: 
+    description: Name part object that represents a Family Name.  There will only ever be a single [NamePart](../NamePartClass) object with a [NamePartType](../NamePartType) of `FamilyName` within the representation of a [Name](../NameClass) for a specific period of validity.  This property will contain the user specified family name or the family name that was read from official documentation.
+    is_a: NamePartClass
+
+slots:
+  type:
+    range: NamePartType
+    required: true
+  nameParts:
+    range: NamePartClass
+    multivalued: true
+    inlined_as_list: true
 
 enums:
   NamePartType:
--- a/v1/linkml-schemas/name.yaml
+++ b/v1/linkml-schemas/name.yaml
@@ -33,11 +33,11 @@ classes:
       - type
 
   GivenNameClass:
-    description: Name part object that represents a Given Name.  There may be multiple [NamePart](../NamePartClass) objects with a [NamePartType](../NamePartType) of `GivenName` within the representation of a [Name](../NameClass) for a specific period of validity.  These will appear in the order in which the user specified or the order in which they were read from official documentation.
+    description: Name part object that represents a Given Name. There may be multiple [NamePart](../NamePartClass) objects with a [NamePartType](../NamePartType) of `GivenName` within the representation of a [Name](../NameClass) for a specific period of validity. These will appear in the order in which the user specified or the order in which they were read from official documentation.
     is_a: NamePartClass
 
   FamilyNameClass: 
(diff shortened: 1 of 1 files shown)


v1/linkml-schemas/name.yaml
```diff
@@ -17,34 +17,37 @@ default_range: string
...
   GivenNameClass:
+    description: Name part object that represents a Given Name.  There may be multiple [NamePart](../NamePartClass) objects with a [NamePartType](../NamePartType) of `GivenName` within the representation of a [Name](../NameClass) for a specific period of validity.  These will appear in the order in which the user specified or the order in which they were read from official documentation.
     is_a: NamePartClass
-    attributes:
-      type:
-        equals_string: GivenName
 
-  FamilyNameClass:
+  FamilyNameClass: 
+    description: Name part object that represents a Family Name.  There will only ever be a single [NamePart](../NamePartClass) object with a [NamePartType](../NamePartType) of `FamilyName` within the representation of a [Name](../NameClass) for a specific period of validity.  This property will contain the user specified family name or the family name that was read from official documentation.
```
galund: This isn't what we're trying to say, I think, let's chat

cmiddi: I agree that this is not in line with RFC11, but this is the way the current implementation is.  There is no code in any of the CRIs to do anything other than return one name part as a family name.  The places where names are currently being given are DCMAW (Passport and Driving License) we know they don't have code to split Family Name Parts.  The other place is in the web journey through the passport CRI and when running this through staging with a space in the family name field we end up with:

```
{
  "sub": "urn:uuid:fb1698db-5675-4eee-a737-c67298b52b93",
  "aud": "https://identity.staging.account.gov.uk",
  "nbf": 1675114802,
  "iss": "https://review-p.staging.account.gov.uk",
  "exp": 1690753202,
  "vc": {
    "evidence": [
      {
        "validityScore": 0,
        "strengthScore": 4,
        "ci": [
          "D02"
        ],
        "txn": "66368711-f788-442a-b637-151c65847447",
        "type": "IdentityCheck"
      }
    ],
    "credentialSubject": {
      "passport": [
        {
          "expiryDate": "2024-09-09",
          "icaoIssuerCode": "GBR",
          "documentNumber": "123456789"
        }
      ],
      "name": [
        {
          "nameParts": [
            {
              "type": "GivenName",
              "value": "JAMES"
            },
            {
              "type": "GivenName",
              "value": "WILLIAM"
            },
            {
              "type": "FamilyName",
              "value": "SMITH WOODS"
            }
          ]
        }
      ],
      "birthDate": [
        {
          "value": "1980-10-01"
        }
      ]
    },
    "type": [
      "VerifiableCredential",
      "IdentityCheckCredential"
    ]
  }
}
```

We need to decide if we think this is tech debt and we should be implementing Family Name splitting to align with RFC11 or we need to change the main wording on RFC11 to bring it in line with the current implementation.  We sort of decided yesterday not to implement Family Name splitting in the DCMAW passport journey as we know the MRZ read can give us problems and it may help us spot them, but it seems like currently there is never going to be any family name splitting anywhere regardless of an MRZ read document or not as it has never been implemented.

cmiddi: Hopefully latest check-in now aligns this to the wording updates on RFC11 and we have reserved the right to use multiple FamilyName name parts in future.

v1/linkml-schemas/name.yaml
```diff
@@ -17,34 +17,38 @@ default_range: string
 
 classes:
   NameClass:
+    description: Name object that represnts a user's claimed identity.  This will have a list of name parts.  Usually this will contain one or more [GivenNameClass](../GivenNameClass) name parts and one [FamilyNameClass](../FamilyNameClass) name part for a specific period of validity.
```
galund:
```suggestion
    description: Name object that represents a user's claimed identity. This will have a list of [`nameParts`](../nameParts). Usually this will contain one or more [`GivenName`](../GivenNameClass) name parts and one (or occasionally more) [`FamilyName`](../FamilyNameClass) name part(s) for a specific period of validity.
```

v1/linkml-schemas/name.yaml
```diff
@@ -17,34 +17,38 @@ default_range: string
...
   GivenNameClass:
+    description: Name part object that represents a Given Name.  There may be multiple [NamePart](../NamePartClass) objects with a [NamePartType](../NamePartType) of `GivenName` within the representation of a [Name](../NameClass) for a specific period of validity.  These will appear in the order in which the user specified or the order in which they were read from official documentation.
     is_a: NamePartClass
-    attributes:
-      type:
-        equals_string: GivenName
 
-  FamilyNameClass:
+  FamilyNameClass: 
+    description: Name part object that represents a Family Name.  There currently will be a single [NamePart](../NamePartClass) object with a [NamePartType](../NamePartType) of `FamilyName` within the representation of a [Name](../NameClass) for a specific period of validity.  This property will contain the user specified family name or the family name that was read from official documentation and may contain space characters.
```
galund: Style thing - would prefer single spacing after full stops :)

v1/linkml-schemas/name.yaml
```diff
@@ -17,34 +17,38 @@ default_range: string
...
     slots:
       - value
-    attributes:
-      type:
-        range: NamePartType
-        multivalued: false
+      - type
+
   GivenNameClass:
+    description: Name part object that represents a Given Name.  There may be multiple [NamePart](../NamePartClass) objects with a [NamePartType](../NamePartType) of `GivenName` within the representation of a [Name](../NameClass) for a specific period of validity.  These will appear in the order in which the user specified or the order in which they were read from official documentation.
```
galund: Technically this may contain space characters too. The example is people called "Anne Marie" where sometimes you get the compound name but without the hyphen. Likely not worth calling out, dunno why I mentioned it really. Not quite sure the level of detail to pitch this documentation!

galund (approved):
I'm basically nitpicking now - am confident we're on the same page!

v1/linkml-schemas/name.yaml
cmiddi: Currently can only happen through DCMAW, but won't once they implement the splitting code, but I get the point.  If you currently type 2 names in first name and 2 names in middle name in Passport CRI you end up with:

`      "name": [
        {
          "nameParts": [
            {
              "type": "GivenName",
              "value": "william"
            },
            {
              "type": "GivenName",
              "value": "james"
            },
            {
              "type": "GivenName",
              "value": "jermaine"
            },
            {
              "type": "GivenName",
              "value": "marcus"
            },
            {
              "type": "FamilyName",
              "value": "Test test test"
            }
          ]
        }
      ]
`

I think if we try to mention it, we could confuse matters more.  Maybe this is where we need our tech writer to help us out :-)

galund: Hah, yep. Most of these cases can only really arise once we give users the magical hypothetical "tell us your name and exactly how you would like us to represent it" user interface... and at least we now have a name for that (the "claimed identity credential issuer", although Kerr doesn't like that name so it may not last :-D )

PR #16, opened 2023-02-07
initial attempt to link to examples from vocab

cmiddi:
Have created some example JSON files from the contents of RFC-0020 and tried to get links from the index and the PostalAddressClass page, a bit crude at the moment, but hopefully we can make it a bit better once we get a tech writer to help out.

Also the build script was complaining about schema.org mappings being explicit in the yaml files, hadn't noticed this before.

### DIFF
--- a/v1/linkml-schemas/address.yaml
+++ b/v1/linkml-schemas/address.yaml
@@ -4,7 +4,6 @@ description: >-
   A postal address associated with a user, returned by an Address Lookup API or by manual data input by the user.
 prefixes:
   linkml: https://w3id.org/linkml/
-  schema: https://schema.org/
   di_vocab: https://vocab.account.gov.uk/v1/
   adb: http://namespaces.geoplace.co.uk/addressbase/2.1
 imports:
@@ -18,7 +17,12 @@ default_range: string
 classes:
   PostalAddressClass:
     description: >-
-      A postal address associated with a user, returned by an Address Lookup API or by manual data input.  
+      A postal address associated with a user, returned by an Address Lookup API or by manual data input.
+      <br />
+      <br />
+      JSON schema - [di_vocab:PostalAddress](json-schemas/PostalAddress.json) 
+      <br />
+      Examples - [Example Postal Address Files](/#postal-address-examples) 
       <br />
       <br />
       Close mapping - [adb:Address](http://www.geoplace.co.uk/addressbase/schema/2.1/addressbase.xsd)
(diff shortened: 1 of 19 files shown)


v1/linkml-schemas/address.yaml
```diff
@@ -18,7 +17,12 @@ default_range: string
 classes:
   PostalAddressClass:
     description: >-
-      A postal address associated with a user, returned by an Address Lookup API or by manual data input.  
+      A postal address associated with a user, returned by an Address Lookup API or by manual data input.
+      <br />
+      <br />
+      JSON schema - [di_vocab:PostalAddress](json-schemas/PostalAddress.json) 
+      <br />
+      Examples - [Example Postal Address Files](/#postal-address-examples)
```
galund: I wonder if it'd be more scalable to have the links to each example here, and have the home page point at this one. Otherwise the home page will end up getting bigger and bigger.

v1/linkml-schemas/person.yaml
```diff
@@ -5,7 +5,6 @@ description: >-
   A real name for a user, composed of an ordered list of given names and family names (etc)
 prefixes:
   linkml: https://w3id.org/linkml/
-  schema: https://schema.org/
```
galund, line 8: We're using the `schema` prefix eg in `birthDate` so I'm surprised this works

cmiddi: Yeah I was surprised but when I have it in the file I get the following when I build: `WARNING:Namespaces:schema namespace is already mapped to http://schema.org/ - Mapping to https://schema.org/ ignored`

galund: I can't figure that out at all, and we may need to come back to it when we look at JSON-LD context creation, but fine then :-)

v1/linkml-schemas/address.yaml
```diff
@@ -18,9 +17,37 @@ default_range: string
...
     description: >-
-      A postal address associated with a user, returned by an Address Lookup API or by manual data input.  
+      A postal address associated with a user, returned by an Address Lookup API or by manual data input.
       <br />
       <br />
+      JSON schema - [di_vocab:PostalAddress](json-schemas/PostalAddress.json) 
+      <br />
+      <br />
+      Examples
+      <ul>
```
galund, line 27: If we can have `<ul>` then we can probably have proper `<p>` markup instead of the `<br>` tags? Sorry this is me getting picky now, this PR is fine ;-)

cmiddi: Yeah I wanted to use actual markup notation but for some reason it doesn't pull through from the yaml files. Let me play with `<br>` and `<p>` and see what I can do although I am sure the accessibility police will come along soon and make it all change anyway!

galund (approved):

PR #17, opened 2023-02-08
format updates to reduce use of br tags

cmiddi:


v1/linkml-schemas/address.yaml
```diff
@@ -17,12 +17,12 @@ default_range: string
...
-      <br />
-      JSON schema - [di_vocab:PostalAddress](json-schemas/PostalAddress.json) 
-      <br />
-      <br />
+      A postal address associated with a user, returned by an Address Lookup API or by manual data input.  
+        
+
+      JSON schema - [di_vocab:PostalAddress](json-schemas/PostalAddress.json)  
+        
+
```
galund: If the bits underneath can be in block elements then I'd imagine this could/should be too?

cmiddi: I will have a look for some reason I could do Markdown new lines here but would be good to make it consistent.

cmiddi: new version pushed - the first block of description cannot be surrounded by tags due to the way it is appended in hierarchical views in other pages.

galund: Urgle, that has to be a bug, but let's roll with the bugs for now :) Can you create log an issue somewhere (in this repo I assume, but if it happened to be straightforward to report as an issue in whatever other repo then do feel free to do that).

galund (approved):

PR #19, opened 2023-02-15
some name examples added

cmiddi:


v1/examples/NameClass/MultipleGivenMultipleFamily.json
```diff
@@ -0,0 +1,20 @@
+{
+    "nameParts": [
+    {
```
galund, line 3: Can I be really annoying and suggest we fix the indents here (we're missing a level for the list)? If you configure the editor with 2 space indents rather than 4, they shouldn't get out of hand ;-)

cmiddi: fixed the list and changed all the examples to have 2 space indentation seemed easier to do that under this pr to keep them consistent :-)

galund (approved):

PR #22, opened 2023-03-09
Changes to document other credentials and introduce a PersonIdentityClass for core attributes as a subset of PersonClass

cmiddi:


galund (approved):
This may break the JSON LD view on things but if it helps the JSON Schemas be correct then that's a win for now.

cmiddi:
Will merge on the premise to fix the schemas, but prepare to iterate if we need to.

PR #27, opened 2023-05-19
Update site links to dirs

cmiddi:
Changes the linkml generator to move classes, enums and slots into their own directories like they do for types.  This then allows for the mkdocs integration to create a better menu structure on the left hand navigation menu.  Much similar to the structure that is given to the overall index document that is generated.  The schema base document that is generated will be generated into a _Home folder to allow for it to appear at the top of the left hand navigation menu generated by mkdocs.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,7 @@
 [submodule "mkdocs-govuk"]
 	path = mkdocs-govuk
 	url = ../mkdocs-govuk.git
-	branch = match-techdocs-template
+	branch = nav_menu_no_root_links
 [submodule "linkml"]
 	path = linkml
 	url = ../linkml.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,4 +5,4 @@
 [submodule "linkml"]
 	path = linkml
 	url = ../linkml.git
-	branch = allow-no-metadata-in-json-schema
+	branch = generate-site-to-dirs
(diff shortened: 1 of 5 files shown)


cmiddi:
Should enable the left hand menu to look something like [this](https://govukverify.atlassian.net/wiki/spaces/~63170c3262fe1e6eac6d6c75/pages/3591668029/DI+Vocabulary)

galund:
I think this is fine tho I'd prefer if the new directory structure was triggered via a command line option, as that would be more likely to ever get accepted by linkml.

I think we should block out some time to go through our changes to linkml and get them to a point where we can make some pull requests, to at least have a plan to move away from being on our own fork one day.

Do we know why our deployment failed?

cmiddi:
Yes the command line trigger needs to be worked out, so it can be triggered as a toggle.  I presumed that the deploy failed as I tried to do it from the branch rather than main, just to work out if it would work and was then going to revert it back.  I have built locally and seems to work ok.  The error message seems to back that up - "Branch "update-site-links-to-dirs" is not allowed to deploy to github-pages due to environment protection rules."  If we can approve this for now we can merge and then look to sort out the flags issue next.  I have raised [https://github.com/alphagov/di-identity-vocab/issues/28](https://github.com/alphagov/di-identity-vocab/issues/28) to track.

galund (approved):

galund:
I've hit approve on the assumption that you'll fix the deployment of the slots before deploying (ie slots should be top-level).

PR #31, opened 2023-06-23
updates to include SecurityCheckCredential

cmiddi:
Updated to add SecurityCheckCredential in line with [RFC-041](https://github.com/alphagov/digital-identity-architecture/blob/main/rfc/0041-CI-mitigation-VC.md)

### DIFF
--- a/v1/linkml-schemas/evidence.yaml
+++ b/v1/linkml-schemas/evidence.yaml
@@ -38,6 +38,32 @@ classes:
       - fraudCheck
       - activityFrom
       - identityCheckLevel
+  SecurityCheckClass:
+    attributes:
+      type:
+        equals_string: SecurityCheck
+    slots:
+      - contraIndicator
+  ContraIndicatorClass:
+    slots:
+      - code
+      - issuanceDate
+      - document
+      - mitigation
+      - incompleteMitigation
+    slot_usage:
+      txn:
+        multivalued: true
+  MitigationClass:
+    slots:
+      - code
+      - mitigatingCredential
+  MitigatingCredentialClass:
+    slots:
+      - issuer
+      - validFrom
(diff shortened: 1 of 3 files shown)


v1/linkml-schemas/evidence.yaml
```diff
@@ -38,6 +38,32 @@ classes:
...
+      type:
+        equals_string: SecurityCheck
+    slots:
+      - contraIndicator
+  ContraIndicatorClass:
+    slots:
+      - code
+      - issuanceDate
+      - document
+      - mitigation
```
elenasantiago1, line 52: should be also include txn slots?

v1/linkml-schemas/evidence.yaml
```diff
@@ -38,6 +38,32 @@ classes:
...
+  MitigationClass:
+    slots:
+      - code
+      - mitigatingCredential
+  MitigatingCredentialClass:
+    slots:
+      - issuer
+      - validFrom
+      - txn
+      - id
```
elenasantiago1, line 66: should we also include: 
incompleteMitigation Class:
slots:
-code
-mitigatingCredential

v1/linkml-schemas/evidence.yaml
cmiddi: txn is slightly below in the slot_usage (line 54), it is because the txn slot is defined at a global level and we are just re-using.

v1/linkml-schemas/evidence.yaml
cmiddi: Incomplete and Complete Mitigations have the same structure so they are both of type MitigationClass

elenasantiago1 (approved):

PR #38, opened 2023-08-10
add devcontainer setup and update readme

olevett:
I hit a couple of issues running through the set up so I thought I'd document them and try and make it simpler for someone in future

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,5 +1,4 @@
 // For format details, see https://aka.ms/devcontainer.json. For config options, see the
-// README at: https://github.com/devcontainers/templates/tree/main/src/anaconda
 {
 	"name": "docs",
 	"image": "mcr.microsoft.com/devcontainers/base:ubuntu",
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,13 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/anaconda
```
galund: I feel like this isn't the correct readme - that's about some specific anaconda dev container (which I'm quite glad to not have to think about as anaconda is a bit weird)

olevett: Interesting, I think I rebuilt this once or twice, I'll just get rid of the readme reference

README.md
```diff
@@ -29,6 +36,10 @@ npm install -g sass  # or your local equivalent per https://sass-lang.com/instal
```
galund, line 36:
```suggestion

If the above dependencies/versions need to change, please update the [devcontainer configuration](.devcontainer/devcontainer.json) also.

```

galund (approved):
(Works for me, just made a suggestion to try and keep the readme and the devcontainer in sync.)

PR #39, opened 2023-08-10
Life events experimental schema

galund:
Add life event for death registration and registration update
- refactored JWT slots into common as we will reuse for security event tokens (SETs)
- refactored Person classes so that we can create schemas that don't include document data for Life Events consumers
- add descriptions as metadata for various elements as required for death notifications
- add "sex" representing legal sex (not for use in One Login)
- add sample death registration and update events

The Life Events work does not publish to the web site yet but arguably this is good because we want to consider it experimental until we have an implementation.


v1/examples/LifeEventSETClass/death-registration.json
```diff
@@ -0,0 +1,80 @@
...
+            "postalCode": "M15 7",
+            "addressCountry": "US"
+          }
+        ]
+      },
+      "deathRegistrationID": 123456,
+      "deathDate": {
+        "value": "1989-07",
+        "description": "Deceased found on 3 August 1989",
+        "verificationLevel": 3
```
olevett: From HMPO, we've heard they don't supply the verificationLevel so we can ignore this for now

olevett:
```suggestion
```

v1/linkml-schemas/lifeEvents.yaml
```diff
@@ -0,0 +1,74 @@
...
+      - freeFormatDeathDate
+  DeathDateClass:
+    mixins:
+      - ValidityClass
+    attributes:
+      value:
+        range: date
+    slots:
+      - description
+      - verificationLevel
```
olevett:
```suggestion
```

v1/linkml-schemas/lifeEvents.yaml
```diff
@@ -0,0 +1,74 @@
...
+    range: DeathDateClass
+    slot_uri: schema:deathDate
+  freeFormatDeathDate:
+    description: Free format date of death of deceased person.
+  deathRegistrationID:
+    description: Registration ID on the RON system
+    range: integer
+  verificationLevel:
+    range: integer
+    description: TBC what this really means
```
olevett:
```suggestion
```

v1/examples/LifeEventSETClass/death-registration.json
```diff
@@ -0,0 +1,79 @@
...
+            "buildingNumber": "5",
+            "streetName": "Mystery Street",
+            "addressLocality": "Coolsville",
+            "postalCode": "M15 7",
+            "addressCountry": "US"
+          }
+        ]
+      },
+      "deathRegistrationID": 123456,
+      "deathDate": {
```
cmiddi: From the document that was shared with DWP by Ethan it had

`"deathDate": [
{
"value": "1989-07-06", # may be partial eg "1989" or "1989-07" or null if freeformDescription is present
"freeformDescription": "On or around the first of October 2023"
}
],`

The `description` as it is modelled here was effectively the option for `freeFormatDeathDate` wasn't it?

v1/examples/LifeEventSETClass/death-registration.json
```diff
@@ -0,0 +1,79 @@
...
+        "value": "1989-07",
+        "description": "Deceased found on 3 August 1989",
+      },
+      "freeFormatDeathDate": "Unknown due to mortuary error but some time in July 1989"
```
cmiddi: As per comment above in Ethan's document this was included under the deathDate as freeformDescription.

v1/linkml-schemas/lifeEvents.yaml
```diff
@@ -0,0 +1,70 @@
...
+      "deathRegistrationUpdate":
+        range: DeathRegistrationEventClass
+        required: true
+        alias: "https://ssf.account.gov.uk/v1/deathRegistrationUpdate"
+  DeathRegistrationEventClass:
+    slots:
+      - deathRegistrationID
+      - subject
+      - deathDate
+      - freeFormatDeathDate
```
cmiddi, line 52:
```suggestion
      - freeFormatDeathDate
      - deathRegistrationDate
```

v1/linkml-schemas/lifeEvents.yaml
```diff
@@ -0,0 +1,70 @@
...
+  DeathDateClass:
+    mixins:
+      - ValidityClass
+    attributes:
+      value:
+        range: date
+    slots:
+      - description
```
cmiddi:
```suggestion
      - description
      
DeathRegistrationDateClass:
  mixins:
    - ValidityClass
  attributes:
    value:
      range: date
 
```

v1/linkml-schemas/lifeEvents.yaml
```diff
@@ -0,0 +1,70 @@
...
+  subject:
+    range: PersonClass
+  deathDate:
+    range: DeathDateClass
+    slot_uri: schema:deathDate
+  freeFormatDeathDate:
+    description: Free format date of death of deceased person.
+  deathRegistrationID:
+    description: Registration ID on the RON system
+    range: integer
```
cmiddi:
```suggestion
    range: integer
  deathRegistrationDate:
    description: Date the death was registered
    range: DeathRegistrationDateClass:
    multivalued: true
    inlined_as_list: true
```

cmiddi (commented):
Initial comments.

v1/examples/LifeEventSETClass/death-registration.json
olevett: From some further discussion with HMPO, we've found two fields that are subtly different
- a qualifier about the death date (e.g.  `Deceased found on`, which is used for partial dates)
- a free format death date, which might be used instead of the date value field and we've got no specific description of what it could be

which we've split into the "qualifier" being the "description", and the free format date being an additional field

cmiddi: Ok thanks.

galund: We definitely need to find a way to explain this in descriptions of the slots, it's quite subtle and I'm not sure I'd really believe what's in there until we get to look at the live event file (assuming someone allows us to do so!!)

v1/linkml-schemas/lifeEvents.yaml
galund: So we reckon that "death registration date" is covered by the SET property `toe` (for death registrations, not for updates, where "death registration date" isn't included anyway.

I think we will need to put a custom description on the `toe` slot/attribute for each event we define - I believe this is possible but haven't attempted it yet.

galund: Although thinking about it, this is weird because we don't actually know the time, so that would have to be zero'd out. There's definitely a case for breaking it out into a separate field like you're suggesting.

cmiddi: I think there is definitely room for ambiguity if we use `toe` so would prefer a specific field in this case.  It may be possible that the event gets registered at a system datetime that is after the date of registration due to manual data input processes, batch input, legacy reporting mechanisms etc.  So I think that `toe` *could* be seen as the system time the event was registered within the GRO database, but the `deathRegistrationDate` is not guaranteed to be the same and as you say wouldn't have a precise time associated with it.

galund: OK cool good argument, I'll commit these suggestions (and squash them, so expect to need to pull the repo down again :) )

v1/linkml-schemas/lifeEvents.yaml
galund: We wouldn't make this multivalued as for the properties at the *event* level (rather than in the subject) we're proposing that these should be single-valued. Happy to chat through in person - I'll do this as a single-valued thing now and then if we decide against we should change all of them at once.

cmiddi: Yes happy to have it single-valued, was just following the pattern we had used previously for things like date of birth, but I think we are safe enough here to know that there will only ever be one.

v1/examples/life-events/death-registration-update.json
```diff
@@ -0,0 +1,82 @@
...
+        ]
+      },
+      "deathRegistrationID": 123456,
+      "deathDate": {
+        "value": "1989-07",
+        "description": "Deceased found on 3 August 1989"
+      },
+      "freeFormatDeathDate": "Unknown due to mortuary error but some time in July 1989",
+      "recordUpdateTime": {
+        "value": "1989-08-10T18:09:10"
```
olevett, line 77: I think we should have a RecordUpdateReason or similar here (and the schema)

v1/examples/life-events/death-registration-update.json
```diff
@@ -0,0 +1,83 @@
...
+              {
+                "value": "Velma",
+                "type": "GivenName"
+              },
+              {
+                "value": "Molone",
+                "type": "FamilyName"
+              }
+            ],
+            "description": "Name before marriage"
```
ethanmills, line 46: Are we proposing mapping the GRO maiden name field to this? Do we need a way of ascribing machine readable meaning to this description, or will it only be used for manual interventions?

v1/linkml-schemas/lifeEvents.yaml
```diff
@@ -0,0 +1,99 @@
...
+classes:
+  DeathRegistrationJWTClass:
+    is_a: SETClass
+    attributes:
+      events:
+        range: DeathRegistrationEventMappingClass
+        required: true
+  DeathRegistrationEventMappingClass:
+    attributes:
+      "deathRegistrationUpdate":
```
ethanmills: Is this right? I think this should be Event, not Update?

v1/examples/life-events/death-registration-update.json
galund: So I think the _current_ proposal is that we would map to this description but discourage (implicitly?) this from being used except in a manual matching process. But if DWP tell us that they need something machine-readable because they want to map to a specific field then we would probably come up with something else (ISTM this would be an optimisation, but eh).

v1/linkml-schemas/lifeEvents.yaml
galund: Ah you're right, this mistake will never have really noticed because of the alias.

I think just `deathRegistration` as that's consistent with the alias, though it's arguable whether "event" should be on the end of both (if we change it, we'd change the update event too).

ethanmills (approved):

cmiddi (approved):

v1/linkml-schemas/lifeEvents.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - deathRegistrationID
+      - deathDate
+      - freeFormatDeathDate
+      - recordUpdateTime
+      - deathRegistrationUpdateReason
+  DeathRegistrationSubjectClass:
+    is_a: PersonExtendedMatchingClass
+    slot_usage:
+      birthDate:
+        range: ISODateClass
```
oskarwilliams, line 68: This doesn't have a description in the GRO XML schema, shall we just not set it?

galund: We have "DeceasedBirthDate" and friends in the [schema](https://docs.google.com/document/d/1e6QbX8pOCBbuIywTXI5cMXl_U4yYljVk/edit) - and generally birth date is exceedingly useful for matching. If it turns out we don't get it then I think that would be a problem that HMPO need to fix?

olevett: I think it's specifically that the "description" isn't set on the birthDate (there's no "reason it's not a full date" field in the schema. I think we can just not set a value for it in our mapping, and leave the possibility that it has a description value in the schema.

galund: Ah I see, sorry. This is because we're using ISODateClass in two cases because both deathDate and birthDate might be partial.

I guess we could make a new class that excludes `description` but I think it doesn't do much harm leaving it in the schema even if it's unused currently in that context.

oskarwilliams: Happy with that, no harm in keeping it and setting it to null :)

galund: General point that we should write down somewhere - we should be consistent about choosing to not include JSON properties at all, vs including them but setting them to null.

Our consistent choice has been to just not include such properties at all (which should make it easier for consumers as there's only way to receive 'no value'). We should do some testing to check whether the schema enforces this at all...

galund: Can confirm the schema validation won't allow `null` - it will allow empty string, but don't do that either :-)

galund: 57604d6251c1e4fa700377f401ed05d15da2a8aa removes `description` from `birthDate` on the example files.

PR #40, opened 2023-08-11
add custom date structure for death date

olevett:


olevett:
In the GRO schema, this formatting applies to both the death date and the birth date

galund (approved):

galund:
I think we should potentially find a way to override the `birthDate` but only for life events - we should probably subclass `Person`.

PR #46, opened 2023-09-11
Simplify creation/update time fields

galund:
- more structure here does not seem particularly valuable
- erring on the side of simplicity can also be a consistency in itself, maybe!

### DIFF
--- a/v1/examples/life-events/death-registered.json
+++ b/v1/examples/life-events/death-registered.json
@@ -73,9 +73,7 @@
         "description": "Deceased found on 3 August 1989"
       },
       "freeFormatDeathDate": "Unknown due to mortuary error but some time in July 1989",
-      "deathRegistrationTime": {
-        "value": "1989-08-10T10:22:04"
-      }
+      "deathRegistrationTime": "1989-08-10T10:22:04"
     }
   }
 }
--- a/v1/linkml-schemas/lifeEvents.yaml
+++ b/v1/linkml-schemas/lifeEvents.yaml
@@ -82,10 +82,10 @@ slots:
     required: true
   deathRegistrationTime:
     description: Date/time the death was registered, may be the same as the `toe` JWT claim
-    range: StructuredDateTimeClass
+    range: datetime
   recordUpdateTime:
     description: Date/time the record was amended, may be the same as the `toe` JWT claim
-    range: StructuredDateTimeClass
+    range: datetime
   deathRegistrationUpdateReason:
     description: A code noting the reason for the update to the death record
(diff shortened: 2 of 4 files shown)


v1/examples/life-events/death-registered.json
```diff
@@ -73,9 +73,7 @@
         "description": "Deceased found on 3 August 1989"
       },
       "freeFormatDeathDate": "Unknown due to mortuary error but some time in July 1989",
-      "deathRegistrationTime": {
-        "value": "1989-08-10T10:22:04"
-      }
+      "deathRegistrationTime": "1989-08-10T10:22:04"
```
ethanmills, line 76: An aside, but we should figure out what timezone this is when provided by GRO. I'd want to convert to Zulu time I think

ethanmills (approved):

v1/linkml-schemas/lifeEvents.yaml
```diff
@@ -82,10 +82,10 @@ slots:
     required: true
   deathRegistrationTime:
     description: Date/time the death was registered, may be the same as the `toe` JWT claim
-    range: StructuredDateTimeClass
+    range: datetime
```
CarlyG55, line 85: It looks like we created `StructuredDateTimeClass`, and we're the only ones using it, so we should probably get rid of that as well?

CarlyG55 (changes requested):

v1/linkml-schemas/lifeEvents.yaml
galund: Good point - fixed - thanks

v1/examples/life-events/death-registered.json
galund: Yeah we ought to call this out in the RFC, because even if we allow `deathRegistrationTime` to be local time, `toe` is still Zulu so we definitely have to do some conversion.

CarlyG55 (approved):

v1/examples/life-events/death-registered.json
cmiddi: I agree that a date generated by a system should be reported in UTC, this aligns with properties of the JWT like `toe`. I think this helps in incident resolution where we may have to look at different component logs running across different timezones / data centre locations.  If consumers want to translate to their local time zone for display purposes then that can be down to them.

galund: Proposal in https://github.com/alphagov/di-identity-vocab/pull/47

olevett (approved):

cmiddi (approved):

PR #47, opened 2023-09-12
Time zones for events

galund:
Follow on from #46 

Specify that times zones are mandatory and should be UTC

- there doesn't seem to be an easy way of enforcing this in JSON Schema
- we might look into writing/finding a regex I suppose

### DIFF
--- a/v1/examples/life-events/death-registered.json
+++ b/v1/examples/life-events/death-registered.json
@@ -73,7 +73,7 @@
         "description": "Deceased found on 3 August 1989"
       },
       "freeFormatDeathDate": "Unknown due to mortuary error but some time in July 1989",
-      "deathRegistrationTime": "1989-08-10T10:22:04"
+      "deathRegistrationTime": "1989-08-10T10:22:04Z"
     }
   }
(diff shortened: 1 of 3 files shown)


ethanmills:
@galund or @olevett do you happen to know how the times come in from GRO source data? Do we need to make code changes to achieve this?

galund:
> @galund or @olevett do you happen to know how the times come in from GRO source data? Do we need to make code changes to achieve this?

No - I've added this to the list of questions we have for HMPO - https://docs.google.com/document/d/1husL1NlFwXd3oqyoI2AX2wtrNiLi0bYcNxgNRy1k84Y/edit#heading=h.g9f7d9bqb2zi

olevett:
The best I've got as an example is 
```xml
<RecordUpdateDateTime>2007-03-06T11:10:01</RecordUpdateDateTime>
```
but with caveats that that's from the DWP example that may or may not be correct

cmiddi (approved):

galund:
I merged this to the wrong bramch...

PR #49, opened 2023-09-14
docs: adds example project.

### LINKED ISSUE #71: Generate TypeScript types from the vocab schemas
Teams maintain their own types that reflect the data model of the vocab schema.

This can lead to inconsistency between the canonical model (vocab) and the implementation, which can lead to issues.

The vocab project should offer types that teams can consume as a build-time asset. These types can be checked against the implementation by the TypeScript transpiler to enable inconsistencies to be detected earlier.

pcornish:
Adds documentation and an example project showing how to consume the TypeScript types.

### Details

This Node.js project consumes the generated TypeScript types.

In the `package.json`, note the `@govuk-one-login/data-vocab` dependency:

```json
"@govuk-one-login/data-vocab": "1.4.2"
```

### Test it

Build the project using:

```shell
npm run build
```

This runs the TypeScript transpiler and outputs to the `dist` directory.

### Run the project

Once you have built the project, run it with:

```shell
npm start
```

You should see a JSON-formatted console entry of a sample `Credentials` object.


galund:
I have it on my list to have a play with this, enough minimally to review and get it landed. I think the first place this may be useful is in TxMA (see <https://govukverify.atlassian.net/browse/DIAR-426>) but it might be something we could consider for Life Events (tho any work there is subject to deprioritisation).

examples/typescript-nodejs/README.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+> **Note**
+> See [releases](https://github.com/govuk-one-login/data-vocab/releases) for the latest version.
+
+### Build the project
+
+##### Prerequisities
+
+This project uses TypeScript and expects the `tsc` command to be available.
```
galund: So I can see that once I've npm-installed `tsc` is available by running `./node_modules/typescript/bin/tsc` ...

Is it normal to also install it globally? Can I get confusing effects if I have a different global tsc to the one I've installed locally?

I could run the npm build / start scripts without installing tsc globally, anyway.

galund (approved):
This works for me and I think it's great. The only question now is that most of the teams are using Java (citation required), so we really need an example project for that :-D

examples/typescript-nodejs/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+Ensure you have authenticated to the `@alphagov` npm scope in GitHub Packages:
+
+```shell
+npm login --scope=@alphagov --auth-type=legacy --registry=https://npm.pkg.github.com
```
jkunle: Is this still `@alphagov` or `gov.uk-one-login` for those who are not familiar using github repos for `npm` I think pre-requisite for this to work should be mentioned. PAT and the scopes required would be helpful

examples/typescript-nodejs/README.md
pcornish: Oh, that's perfect - I'll update the docs :)

pcornish:
> @galund This works for me and I think it's great. The only question now is that most of the teams are using Java (citation required), so we really need an example project for that :-D

This is a great suggestion. Raised #72 to track that work.

examples/typescript-nodejs/README.md
pcornish: Thank you for catching this - updated to the new org name 👍

Issue #42, opened 2023-09-06
Links to schemas too prominent


galund:
It's easy to see a link to "name" on the home page click it, only to be plunged straight into a JSON Schema document.

This isn't user friendly - JSON schemas need relegating elsewhere, linked to from the relevant human-readable pages.

chris-dco:
This ticket has gone stale. Please open a new ticket if further changes are required.

PR #63, opened 2023-10-19
Remove schema files from home page

galund:
- JSON schemas are not the most user-friendly way into discovering the Vocabulary
- instead, generate an index file containing all of the JSON schemas that we build
- also, ensure they are linked to from each relevant class
- for now at least, link both in the description and with see_also, consistently

Fixes issue https://github.com/govuk-one-login/data-vocab/issues/42

Screenshot of new JSON Schemas index page

### DIFF
--- a/scripts/generate_json_schemas.sh
+++ b/scripts/generate_json_schemas.sh
@@ -11,22 +11,22 @@ set -e
 
 # format: <linkml schema file>,<linkml class>,<json schema file>
 LINKML_ITEMS=(
-  "credentials.yaml,CoreIdentityJWTClass,CoreIdentityJWT.json"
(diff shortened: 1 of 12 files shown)


scripts/generate_json_schemas.sh
```diff
@@ -13,9 +13,9 @@ set -e
 LINKML_ITEMS=(
   "credentials.yaml,CoreIdentityJWTClass,CoreIdentityJWT.json"
```
galund: Is there a good ordering for these, now that the ordering shows up in the index file?

Would alphabetical be better than the slightly-arbitrary order we have now?

cmiddi: I think alphabetic is better and I am sure the tech writer and accessibility community will agree :-)

cmiddi (approved):

PR #69, opened 2023-10-27
SPT-482:update Vocab schema to add extra validation

chaitanguttarlapalli:
**What:**

Differences have been identified between the SPOT MVP credential schema and the schema published in the [vocab repo]. Update the vocab site, to reflect the additions as detailed in the acceptance criteria.

**Why:**
The schema published in the vocab repo should reflect the SPOT checks, such that any CRI checking their work against schema will not be caught by any additional validation enforced by the SPOT schema.
We are updating the vocab schema so it becomes the source of truth and, ultimately, will be using the vocab schema in SPOT instead of the current MVP schema.

**AcceptanceCriteria:**
The following changes to the schema have been made:
$ - Should have the following required claims sub, iss, nbf, and vc
$.jti - Should be a non-empty string uri
$.vtm - Should be a non-empty string uri
$.vc - Should have evidence and type required
$.vc["@context"] -Should be a non-empty string
$.vc.evidence[] - Should require a type and txn
$.vc.evidence[].activityHistoryScore - Should have a max 4
$.vc.evidence[].ci[] - Should support either null or string
$.vc.evidence[].identityFraudScore -Should have a max max 3
$.vc.evidence[].strengthScore -Should have a max 4
$.vc.evidence[].txn - Should support string and string[]
$.vc.evidence[].validityScore -Should have a max 4
$.vc.evidence[].verificationScore - Should have a max 4
$.vc.credentialSubject[].birthDate.value - Should be required
$.vc.credentialSubject[].name[].nameParts - Should be required

### DIFF
--- a/v1/examples/life-events/death-registered.json
+++ b/v1/examples/life-events/death-registered.json
@@ -2,8 +2,6 @@
   "iss": "https://life-events-platform.gov.uk.example",
   "iat": 1691637001,
   "jti": "4d3559ec67504aaba65d40b0363faad8",
-  "sub": "urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6",
-  "nbf": 1691637001,
   "toe": 618747724,
   "events": {
     "https://vocab.account.gov.uk/v1/deathRegistered": {
--- a/v1/examples/life-events/death-registered.json
+++ b/v1/examples/life-events/death-registered.json
@@ -2,6 +2,8 @@
   "iss": "https://life-events-platform.gov.uk.example",
   "iat": 1691637001,
   "jti": "4d3559ec67504aaba65d40b0363faad8",
+  "sub": "urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6",
+  "nbf": 1691637001,
   "toe": 618747724,
   "events": {
     "https://vocab.account.gov.uk/v1/deathRegistered": {
--- a/v1/linkml-schemas/common.yaml
+++ b/v1/linkml-schemas/common.yaml
@@ -74,12 +74,15 @@ slots:
     range: date
     slot_uri: schema:validUntil
   iss:
+    required: true
     range: uri
     description: The `iss` (issuer) claim identifies the principal that issued the JWT.
   sub:
+    required: true
     range: uri
     description: The `sub` (subject) claim identifies the principal that is the subject of the JWT.
   nbf:
+    required: true
     range: integer
     description: The `nbf` (not before) claim identifies the time the JWT is valid from.
   exp:
--- a/v1/linkml-schemas/common.yaml
+++ b/v1/linkml-schemas/common.yaml
@@ -74,15 +74,12 @@ slots:
     range: date
     slot_uri: schema:validUntil
   iss:
-    required: true
     range: uri
     description: The `iss` (issuer) claim identifies the principal that issued the JWT.
   sub:
-    required: true
     range: uri
     description: The `sub` (subject) claim identifies the principal that is the subject of the JWT.
   nbf:
-    required: true
     range: integer
     description: The `nbf` (not before) claim identifies the time the JWT is valid from.
   exp:
--- a/v1/linkml-schemas/credentials.yaml
+++ b/v1/linkml-schemas/credentials.yaml
@@ -93,10 +93,24 @@ classes:
     see_also:
       - ../json-schemas/IdentityCheckCredentialJWT.json
     attributes:
+      iss:
+        range: uri
+        description: The `iss` (issuer) claim identifies the principal that issued the JWT.
+        required: true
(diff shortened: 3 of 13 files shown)


v1/examples/life-events/death-registered.json
```diff
@@ -2,6 +2,8 @@
   "iss": "https://life-events-platform.gov.uk.example",
   "iat": 1691637001,
   "jti": "4d3559ec67504aaba65d40b0363faad8",
+  "sub": "urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6",
```
galund: These events do not have subject identifiers so this is not possible.

v1/linkml-schemas/common.yaml
```diff
@@ -74,12 +74,15 @@ slots:
...
   iss:
+    required: true
     range: uri
     description: The `iss` (issuer) claim identifies the principal that issued the JWT.
   sub:
+    required: true
     range: uri
     description: The `sub` (subject) claim identifies the principal that is the subject of the JWT.
   nbf:
+    required: true
```
galund: In general whether a particular JWT claim is required or not is a very much context-dependent. So I think this should be in slot_usage or similar.

This should follow RFC 7519 which specifically says
> None of the claims defined below are intended to be mandatory to use or implement in all cases, but rather they provide a starting point for a set of useful, interoperable claims.

galund (changes requested):
This isn't quite right, you can't just add data to examples and expect it to be correct. If the examples don't contain something, it may be for a reason :)

v1/linkml-schemas/evidence.yaml
```diff
@@ -122,6 +122,7 @@ enums:
 slots:
   txn:
     description: A unique transaction identifier for this check, or part of a check, if any.
+    range: string
```
galund, line 125: this is the default anyway

v1/linkml-schemas/credentials.yaml
```diff
@@ -93,10 +93,24 @@ classes:
     see_also:
       - ../json-schemas/IdentityCheckCredentialJWT.json
     attributes:
+      iss:
+        range: uri
+        description: The `iss` (issuer) claim identifies the principal that issued the JWT.
```
chris-dco, line 98: The build does not like the ` character. Replace it with ' or remove it.

v1/linkml-schemas/credentials.yaml
```diff
@@ -93,10 +93,24 @@ classes:
...
+        required: true
+      sub:
+        range: uri
+        description: The `sub` (subject) claim identifies  jaja the principal that is the subject of the JWT.
```
chris-dco, line 102: The build does not like the ` character. Replace it with ' or remove it.

v1/linkml-schemas/credentials.yaml
```diff
@@ -93,10 +93,24 @@ classes:
...
+        required: true
+      nbf:
+        range: integer
+        description: The `nbf` (not before) claim identifies the time the JWT is valid from.
```
chris-dco, line 106: The build does not like the ` character. Replace it with ' or remove it.

v1/linkml-schemas/credentials.yaml
```diff
@@ -93,10 +93,24 @@ classes:
...
+        required: true
       vc:
         range: IdentityCheckCredentialClass
         required: true
-
+      jti:
+        range: uri
+        pattern: "^(?!\s*$).+"
```
chris-dco, line 113: The build does not like the " character so you'll need to remove it.

v1/linkml-schemas/name.yaml
```diff
@@ -55,18 +55,26 @@ classes:
     abstract: false  # has to be false otherwise JSON-Schema doesn't include it, and validation fails :(
     mixins:
       - ValidityClass
-    slots:
+    slot:
```
chris-dco, line 58: This should be `slots`.

v1/linkml-schemas/vectorsOfTrust.yaml
```diff
@@ -27,3 +27,4 @@ slots:
     range: IdentityVectorOfTrust
   vtm:
     range: uri
+    description: The `vtm` identifies the Vector Trust Mark.
```
chris-dco, line 30: The build does not like the ` character. Replace it with ' or remove it.

v1/linkml-schemas/credentials.yaml
galund: Hmm I'd have thought the quote mark here is a YAML delimiter and so never makes it through to the code, this seems weird

cmiddi:
Do we want to abandon this as I think it has been replaced by [this one](https://github.com/govuk-one-login/data-vocab/pull/73)

PR #70, opened 2023-11-01
dataCheck property

cmiddi:
Adding a new property dataCheck to the CheckDetailsClass and restricting the values to new Enum type DataCheckType.

### DIFF
--- a/v1/linkml-schemas/evidence.yaml
+++ b/v1/linkml-schemas/evidence.yaml
@@ -121,7 +121,7 @@ enums:
         impersonation_risk_check:
   DataCheckType:
       permissible_values:
-        data_verification_check:
+        verification_check:
         cancelled_check:
         lost_stolen_check:
 
--- a/v1/linkml-schemas/evidence.yaml
+++ b/v1/linkml-schemas/evidence.yaml
@@ -121,7 +121,7 @@ enums:
         impersonation_risk_check:
   DataCheckType:
       permissible_values:
-        verification_check:
+        record_check:
         cancelled_check:
         lost_stolen_check:
 
--- a/v1/linkml-schemas/evidence.yaml
+++ b/v1/linkml-schemas/evidence.yaml
@@ -191,7 +191,7 @@ slots:
     description: Specifies the kind of fraud check performed as defined in the [Good Practice Guide 45 documentation](https://www.gov.uk/government/publications/identity-proofing-and-verification-of-an-individual/how-to-prove-and-verify-someones-identity#identity-fraud)
   dataCheck:
     range: DataCheckType
-    description: Specifies the kind of data check performed allows for the distinction between a data verification, cancelled check and a lost or stolen check
+    description: Specifies the kind of data check performed allows for the distinction between a record check, cancelled check and a lost or stolen check
   activityFrom:
     range: date
     description: The date of the earliest activity found for the user.
--- a/v1/linkml-schemas/evidence.yaml
+++ b/v1/linkml-schemas/evidence.yaml
@@ -36,6 +36,7 @@ classes:
       - photoVerificationProcessLevel
       - biometricVerificationProcessLevel
       - fraudCheck
+      - dataCheck
       - activityFrom
       - identityCheckLevel
       - identityCheckPolicy
@@ -118,6 +119,11 @@ enums:
         identity_theft_check:
         synthetic_identity_check:
         impersonation_risk_check:
+  DataCheckType:
+      permissible_values:
+        data_verification_check:
+        cancelled_check:
+        lost_stolen_check:
 
 slots:
   txn:
@@ -183,6 +189,9 @@ slots:
   fraudCheck:
     range: FraudCheckType
     description: Specifies the kind of fraud check performed as defined in the [Good Practice Guide 45 documentation](https://www.gov.uk/government/publications/identity-proofing-and-verification-of-an-individual/how-to-prove-and-verify-someones-identity#identity-fraud)
+  dataCheck:
+    range: DataCheckType
+    description: Specifies the kind of data check performed allows for the distinction between a data verification, cancelled check and a lost or stolen check
   activityFrom:
     range: date
     description: The date of the earliest activity found for the user.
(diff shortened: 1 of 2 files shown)


galund:
I'm kinda interested to know why this is required, does it link with an update to RFC 24?

If a credential issuer supports checking its database at all, and a piece of evidence is capable of being recorded as lost or stolen or cancelled, then you'd expect that the check would include that. I don't really see why we'd record that into the VC.

'Verification check' is confusing - do you mean 'validity check'? Again I think all checks of this sort validate some piece of evidence, digital or otherwise. But verification is something completely else.

cmiddi:
> I'm kinda interested to know why this is required, does it link with an update to RFC 24?
> 
> If a credential issuer supports checking its database at all, and a piece of evidence is capable of being recorded as lost or stolen or cancelled, then you'd expect that the check would include that. I don't really see why we'd record that into the VC.
> 
> 'Verification check' is confusing - do you mean 'validity check'? Again I think all checks of this sort validate some piece of evidence, digital or otherwise. But verification is something completely else.

Yes links to an update in RFC-0024 there is a pull request out to update that in line.  So we had a session where we were discussing the direct integration with HMPO following the demise of DCS and Emily alerted us to the fact that thew new DVA Digital interface was actually doing potentially 3 checks behind the scenes and we were getting returned this information as a series of flags.  The first check is a verification check to ensure that the document that was specified exists within the database and is appropriated to the correct person details specified.  The second check was a cancelled check to inform if the passport has been cancelled by HMPO for any reason.  The third check is a lost or stolen check to see if it is on the list of lost and stolen passports (all passports that are reported lost and stolen are automatically cancelled).  The idea behind putting these details in the VC would allow us to see which of the checks were run as per the pattern in the fraudCheck property.  It would also allow us to start using a `D01` and let us distinguish if a VC had a `D01` for being cancelled or also lost or stolen.  It also allows the fraud team to do greater analysis.

galund:
> > I'm kinda interested to know why this is required, does it link with an update to RFC 24?
> > If a credential issuer supports checking its database at all, and a piece of evidence is capable of being recorded as lost or stolen or cancelled, then you'd expect that the check would include that. I don't really see why we'd record that into the VC.
> > 'Verification check' is confusing - do you mean 'validity check'? Again I think all checks of this sort validate some piece of evidence, digital or otherwise. But verification is something completely else.
> 
> Yes links to an update in RFC-0024 there is a pull request out to update that in line. So we had a session where we were discussing the direct integration with HMPO following the demise of DCS and Emily alerted us to the fact that thew new DVA Digital interface was actually doing potentially 3 checks behind the scenes and we were getting returned this information as a series of flags. The first check is a verification check to ensure that the document that was specified exists within the database and is appropriated to the correct person details specified. The second check was a cancelled check to inform if the passport has been cancelled by HMPO for any reason. The third check is a lost or stolen check to see if it is on the list of lost and stolen passports (all passports that are reported lost and stolen are automatically cancelled). The idea behind putting these details in the VC would allow us to see which of the checks were run as per the pattern in the fraudCheck property. It would also allow us to start using a `D01` and let us distinguish if a VC had a `D01` for being cancelled or also lost or stolen. It also allows the fraud team to do greater analysis.

FWIW I don't think there should be any limit on which credential issuers can issue which contra-indicators, so I'd not see this as a pre-req for using D01. Even if you add this, you can't tell which one generated the D01, so if you need to distinguish then you need different contra-indicator codes.

galund:
I guess another point with this is that we added the extra detail to the fraud check because it's exactly enough detail to help you determine the "risk of identity fraud" score. And in particular because we might want to arrive at a fraud score by checking with more than one credential issuer. But the three things in this PR aren't enough to let a VC verifier determine the validity score. Validity scores are a lot more complicated, and my advice would be to stick with the scoring rather than try to pull this out, because a given credential issuer should always be able to determine the score on its own, following those complicated rules.

[edit: argh, validity not verification! it's so easy to confuse things :) ]

cmiddi:
> Even if you add this, you can't tell which one generated the D01, so if you need to distinguish then you need different contra-indicator codes.

You will be able to tell based on which checks appear in the checkDetails and failedCheckDetails, as there will be an item for each check that has taken place.  If a passport is lost or stolen and cancelled there will be 2 elements in failedCheckDetails, if it is not lost or stolen but just cancelled there will be 1 element and 2 elements in checkDetails.  Sudha's [confluence page](https://govukverify.atlassian.net/wiki/spaces/Architecture/pages/3590553679/Passport+CRI+Technical+Specification#Example-VCs) gives more details.  The associated RFC PR is [here](https://github.com/govuk-one-login/architecture/pull/558).

v1/linkml-schemas/evidence.yaml
```diff
@@ -183,6 +189,9 @@ slots:
   fraudCheck:
     range: FraudCheckType
     description: Specifies the kind of fraud check performed as defined in the [Good Practice Guide 45 documentation](https://www.gov.uk/government/publications/identity-proofing-and-verification-of-an-individual/how-to-prove-and-verify-someones-identity#identity-fraud)
+  dataCheck:
+    range: DataCheckType
+    description: Specifies the kind of data check performed allows for the distinction between a data verification, cancelled check and a lost or stolen check
```
cmiddi:
```suggestion
    description: Specifies the kind of data check performed allows for the distinction between a record check, cancelled check and a lost or stolen check
```

v1/linkml-schemas/evidence.yaml
```diff
@@ -118,6 +119,11 @@ enums:
         identity_theft_check:
         synthetic_identity_check:
         impersonation_risk_check:
+  DataCheckType:
+      permissible_values:
+        verification_check:
```
cmiddi:
```suggestion
        record_check:
```

cmiddi:
The intention behind this is not to change or affect the scoring in any way it is a means to provide more granularity around the nature of the checks that have been carried out.  This has benefits outside the identity proving process and allows us to start using alternative contra-indicator codes that are more appropriate to the check failure seen.  The use of verification in the data check type was a bad choice of words when trying to deviate away from the word valid for something that may be also cancelled.  Having discussed this to remove confusion the proposal is simply to go for something away from valid / verify and use simply `record_check`.

smohan-gds (approved):

cmiddi:
Merging in line with RFC after agreement at tech leads meeting on 10/11.

Issue #54, opened 2023-10-16
Problems in table of contents


galund:
There are a few things that are broken about the table of contents:

- the 'V1' link is the root of all TOC items and links to a completely obscure `@context` property that probably shouldn't even be in the metamodel. (It's only there until we can figure out how to generate JSON Schemas that allow JSON-LD `@context` properties some other way.)
- the 'GOV.UK Vocabulary' item at the top should probably link to the home page ie duplicate the 'Vocabulary' link in the top menu bar (but see <https://github.com/govuk-one-login/data-vocab/issues/21>).
- the home/credentials item in the TOC is a bit strange, though have a full contents _somewhere_ is probably a good thing.

GDS-Dominic:
Top two issues have been resolved with new theme however we need to determine do we want a full schema (home/credentials item) - Its kinda nice but also i'm really not sure how many people who would use it

galund:
https://github.com/govuk-one-login/data-vocab/pull/76 ended up being more radical than I expected, but ... hurrah for deleting code :-) ?

chris-dco:
This ticket has gone stale. Please open a new ticket if further changes are required.

PR #76, opened 2023-11-21
Replace custom script with existing MkDocs plugin

galund:
Simplification, along with fix for https://github.com/govuk-one-login/data-vocab/issues/54

The plugin has some issues, but this does provide a fairly simple fix for the issues in the menu.

There's still a duplication of the site name at the top of the menu (which looks odd at some screen resolutions only!) but this is maybe good enough.


GDS-Dominic:
I am wary about this whilst less code could be quicker and easier to maintain I think current script provides us some useful flexibility i.e. in my current branch for audit events i've been able to do a lot more manipulation with it which has come in useful
Screenshot 2023-11-22 at 11 23 43

galund:
I *think* we can build a structure like that pretty easily... which branch, I don't mind having a play?

If we do decide we want something custom then I think it might be worth writing it into an extension rather than overwriting the nav file - and I started looking at that before realising that someone had in fact already built something that does the job.

Joe-Edwards-GDS:
Superseded by #147

PR #80, opened 2024-01-16
feat: adds Java type generation.

### LINKED ISSUE #72: Generate Java data model classes from the vocab schemas
Teams maintain their own data model classes that reflect the data model of the vocab schema.

This can lead to inconsistency between the canonical model (vocab) and the implementation, which can lead to issues.

The vocab project should offer data model classes that teams can consume as a build-time asset. These types can be checked against the implementation by the Java compiler to enable inconsistencies to be detected earlier.

pcornish:
fixes #72 

## What does this do

This PR adds a project that generates data model classes as Java POJOs, based on the JSON schema files.

The generated Java types are written to the `build/generated-sources/di-data-model` directory.

### Usage

Here is an example using the `IdentityCheckCredentialJWT` class.

> **Note**
> In this example, we make use of the Gang-of-Four builder pattern (the static `builder()` method, and its
> associated `with...` methods).
> 
> If preferred, the POJO's constructor, and its setters and getters can be used to construct
> instances instead. 

```java
var credentials = IdentityCheckCredentialJWT.builder()
    .withSub("urn:fdc:gov.uk:2022:954bc117-731b-41cd-86cf-dfb4e7940fce")
    .withAud("https://passport.core.stubs.account.gov.uk")
    .withNbf(1690816091)
    .withIss("https://review-p.build.account.gov.uk")
    .build();

var vc = IdentityCheckCredentialClass().builder()
    .withType(List.of(
        VerifiableCredentialType.VERIFIABLE_CREDENTIAL,
        VerifiableCredentialType.IDENTITY_CHECK_CREDENTIAL
    )).build();

credentials.setVc(vc);

var evidence = IdentityCheckClass.builder()
    .withValidityScore(0)
    .withStrengthScore(4)
    .withCi(List.of("D02"))
    .withTxn("5f57a8f2-62b0-4958-9332-06d9f453e5b9")
    .withType("IdentityCheck")
    .build();

vc.setEvidence(List.of(evidence));
```

## Example

See the unit test `uk.gov.di.model.ModelTest` (under `src/test/java`) for an example of using the
model, then serialising it to JSON.

### Prerequisites

- Generate the schema files from this project (see `scripts/generate_json_schemas.sh`)
- Ensure you have Java 17+ installed

## Steps

Generate the Java types and run the unit tests:

```shell
./gradlew build
```

The generated Java types are written to the `build/generated-sources/di-data-model` directory.

## Implementation notes

The class generation is done using the widely-used https://github.com/joelittlejohn/jsonschema2pojo project (specifically its Gradle plugin).

The model classes are annotated using Jackson (though other options are possible).

## Publishing

Like the TypeScript variant, the build pipeline produces and publishes an artifact consumable by dependency management tools.

The Maven artifact is published here: https://github.com/orgs/govuk-one-login/packages?repo_name=data-vocab

The Maven publishing configuration is [documented here](https://github.com/govuk-one-login/data-vocab/blob/feat/java-types/scripts/java-types/README.md#publish-maven-artifact).


fewtrees-pa (commented):
The principle is good with the following caveats:

The Builder pattern in https://github.com/joelittlejohn/jsonschema2pojo doesn’t generate builders but syntactic sugar to circumvent the bean method setter standard and is a bit of a smell.. I personally would not use it and use Lombok Builder annotations instead.

I'm not sure if adding the Json serialisation annotations is a good idea - it could want versions that do not comply/agree with the currently supported versions we use. I've had many headaches from cross-dependencies with json libs..

Will the tool handle multiple input schema files? There's only 1 output package name and each inout schema file also generates a `$defs` class - will it aggregate schema defs

What is really needed is a tool that converts json schema to a lombok annotated class (https://projectlombok.org/features/Builder) so all you need is:

```
@Builder
@Data
class MyObj {
  String firstName;
  String lastName;
}
```

..and then you get everything else for free..

Joe-Edwards-GDS:
@pcornish I'm pretty interested in this, as we've been exploring doing the same thing in IPV Core (using the same library!).

We've only been prototyping so far, and broadly speaking it seems to work pretty well. Having it published here would be even better (and IMO is a better way of tracking schema compatibility than the current iteration of contract testing), so if that's coming soon, we'll avoid duplicating the implementation.

A few issues I encountered while prototyping this:
- You get slightly different results depending on what you choose as the top level (the JWT or Credential class)
    - I guess we should just pick all the distinct JWT classes
- There is duplication when building schemas that define the same internal types - e.g. I got a load of copies of `VerifiableCredentialType`, `NamePart` etc.
- The inheritance structure isn't well reflected, and Java's lack of union types makes it hard to ad-hoc it
    - There are a bunch of things you can do in the json-schemas to customise this, but the simplest might be to add something like `"javaInterfaces": ["uk.gov.di.model.VerifiableCredentialJWT"]` to the JWT schemas.
    - You need to define that interface manually, and ideally it would expose all the shared fields, but just having a shared empty interface is still really useful for consuming code.
    - You can also reference a base class, e.g. with `extendsJavaClass`, but inheritance doesn't work great with pojos
    - Anything cleverer would probably need to be a custom ruleset for jsonschema2pojo
- The type definitions are not all complete
    - Various missing fields and e.g. no `RiskAssessmentCredential`
    - This is really a positive as it's an opportunity to bring these definitions up to date
    - We'll need to take a lot of care to make sure nothing will break in prod :)

On @fewtrees-pa 's comment about serialisation - it feels somewhat necessary to have some annotations, since we're liable to have a few attributes with weird names. However the compatibility point is a sticky one - e.g. at the moment IPV Core can't upgrade to the latest version of jackson because it conflicts with lambda powertools.

danbudden-gds:
This sounds like a really good idea. 

Those of us in the `data-architects` group will need to rely on you @Joe-Edwards-GDS, @pcornish , and @fewtrees-pa to iron out the technical side. If you can get yourselves comfortable where you can approve it? 

It does sound like there may be some modelling work or other aspects of the schema work to do, which `data-architects` can take on.

pcornish:
Updated following conversation with @fewtrees-pa 

There are now two artifacts:
* plain POJO artifact, with no annotations
* annotated artifact, with Jackson annotations

To one of your points @Joe-Edwards-GDS, the version on the Jackson dependency is now specified as a range:

```
com.fasterxml.jackson.core:jackson-databind:[2.15.2,)
```

> [Version syntax docs](https://docs.gradle.org/current/userguide/single_versions.html)

...so as to be permissive to the consuming project's version.

code-generators/java-types/plain-pojo/src/test/resources/serialised_model.json
```diff
@@ -0,0 +1,62 @@
+{
+  "aud" : "https://passport.core.stubs.account.gov.uk",
+  "exp" : null,
+  "iat" : null,
+  "iss" : "https://review-p.build.account.gov.uk",
+  "jti" : null,
+  "nbf" : 1690816091,
+  "sub" : "urn:fdc:gov.uk:2022:954bc117-731b-41cd-86cf-dfb4e7940fce",
+  "vc" : {
+    "context" : null,
```
Joe-Edwards-GDS: Note that in the plain POJO this property name is incorrect - it should be `@context` (not `context`). It doesn't really matter since afaict we never actually populate this property - it's a W3 VC property https://www.w3.org/TR/vc-data-model/#contexts - but a potential gotcha

pcornish: Good spot! This is resolved in the test now. The serialisation wasn't configured properly, but now the serialised JSON is the same between the annotated and plain modules.

PR #81, opened 2024-02-15
SPT-679 SPOT: Update vocab schema

chris-dco:
## What

Update the vocab schema to reflect the fact that the `vot` can be either `PCL200` or `PCL250`.

## Why

Currently, the `vot` is defined as an enum containing values `P1`, `P2`, `P3` and `P4`. This needs reflect that new values are permissible.

## Acceptance Criteria

The following acceptance criteria were not followed. This is due to issues with the currently implementation whereby the `IdentityVectorOfTrust` and the new enum would result in the type being `string`. The SPOT team looked into this and determined that the best course of action would be to add the `PCL200` or `PCL250` to the existing `IdentityVectorOfTrust`. However, later investigations may be carried out into the tooling to determine if this is possible in the future.

* A new enum is defined for inherited identities that allows the values  `PCL200` and `PCL250`.
* The existing `IdentityVectorOfTrust` is not modified and continues to allow only `P1`, `P2`, `P3` and `P4`.
* The definition of the `vot` field is updated to allow a value from the union of the two enums.

### DIFF
--- a/v1/linkml-schemas/vectorsOfTrust.yaml
+++ b/v1/linkml-schemas/vectorsOfTrust.yaml
@@ -10,7 +10,7 @@ imports:
   - linkml:types
   - ./common
 default_curi_maps:
-  - semweb_context  
+  - semweb_context
 default_prefix: di_vocab
 default_range: string
 
@@ -21,10 +21,13 @@ enums:
       P2:
       P3:
       P4:
+      PCL200:
+      PCL250:
 
 slots:
   vot:
     range: IdentityVectorOfTrust
+    description: The `vot` identifies the Vector Of Trust.
   vtm:
     range: uri
     description: The `vtm` identifies the Vector Trust Mark.
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -7,8 +7,10 @@
 		"ghcr.io/devcontainers/features/python:1": {},
 		"ghcr.io/devcontainers-contrib/features/poetry:2": {}
 	},
-	"forwardPorts": ["8000:8000"],
-	"postCreateCommand": "poetry install && npm ci && npm install -g ajv-cli@5.0.0 ajv-formats@2.1.1 sass",
+	"forwardPorts": [
+		"8000:8000"
+	],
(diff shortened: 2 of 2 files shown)


michaelabenyohai:
If I build this branch I am not sure I am seeing what I would expect in `v1/json-schemas/IdentityCheckCredentialJWT.json`. If I build on `main` branch I currently get:

```
"vot": {
               "$ref": "#/$defs/IdentityVectorOfTrust"
 }

...

"IdentityVectorOfTrust": {
         "description": "",
         "enum": [
            "P1",
            "P2",
            "P3",
            "P4"
         ],
         "title": "IdentityVectorOfTrust",
         "type": "string"
}
```

If I build on this branch I get:

```
"vot": {
               "type": "string"
}

...

"IdentityVectorOfTrust": {
         "description": "",
         "enum": [
            "P1",
            "P2",
            "P3",
            "P4"
         ],
         "title": "IdentityVectorOfTrust",
         "type": "string"
 }

...

"OperationalIdentityVectorOfTrust": {
        "description": "",
         "enum": [
            "PCL200",
            "PCL250"
         ],
         "title": "OperationalIdentityVectorOfTrust",
         "type": "string"
}
```

I think I'd expect the `vot` to refer to the `IdentityVectorOfTrust` and `OperationalIdentityVectorOfTrust` types in some way rather than being of type `string`. Is this working as expected?

v1/linkml-schemas/vectorsOfTrust.yaml
```diff
@@ -21,9 +21,12 @@ enums:
       P2:
       P3:
       P4:
+      PCL200:
+      PCL250:
 
 slots:
   vot:
+    description: The `vot` identifies the Vector Of Trust.
```
michaelabenyohai: Sorry to be petty but could this be rearranged to be:

```
range: IdentityVectorOfTrust
description: The `vot` identifies the Vector Of Trust.
```

rather than:

```
description: The `vot` identifies the Vector Of Trust.
range: IdentityVectorOfTrust
```

just to be consistent with other parts of this file and other files :)

chris-dco: @michaelabenyohai Done

michaelabenyohai (approved):

chrisclayson (approved):

cmiddi (approved):

PR #82, opened 2024-02-20
update crown

cmiddi:
Update the favicon and the crown still needs some template work but this will do for now.


docs/assets/govuk/assets/images/govuk-logotype-crown.png
erino, line 1: I think we should use this version of the new crown https://github.com/alphagov/govuk-frontend/blob/v4.8.0/src/govuk/assets/images/govuk-logotype-tudor-crown.png

cmiddi: Agreed, was looking for something like that, but couldn't find it so was going to fix later.  Thanks!

GDS-Dominic (approved):

PR #95, opened 2024-04-04
Adds JWT Classes and corresponding json-schema for other VCs

Joe-Edwards-GDS:
Adds new class for `AddressCredentialJWTClass`

Adds json-schema generation for `AddressCredentialJWT`, `IdentityAssertionCredentialJWT`, `RiskAssessmentCredentialJWT` to bring them in line with other JWTs.

### DIFF
--- a/scripts/generate_json_schemas.sh
+++ b/scripts/generate_json_schemas.sh
@@ -11,12 +11,14 @@ set -e
 
 # format: <linkml schema file>,<linkml class>,<json schema file>
 LINKML_ITEMS=(
+  "credentials.yaml,AddressCredentialJWTClass,AddressCredentialJWT.json"
   "credentials.yaml,AuthorizationRequestClass,AuthorizationRequest.json"
   "credentials.yaml,CoreIdentityJWTClass,CoreIdentityJWT.json"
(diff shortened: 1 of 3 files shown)


scripts/generate_json_schemas.sh
```diff
@@ -27,6 +29,7 @@ LINKML_ITEMS=(
   "address.yaml,PostalAddressClass,PostalAddress.json"
   "document.yaml,ResidencePermitDetailsClass,ResidencePermit.json"
   "riskAssessmentCredential.yaml,RiskAssessmentCredentialClass,RiskAssessmentCredential.json"
+  "credentials.yaml,RiskAssessmentCredentialJWTClass,RiskAssessmentCredentialJWT.json"
   "securityCheckCredential.yaml,SecurityCheckCredentialClass,SecurityCheckCredential.json"
```
Joe-Edwards-GDS, line 33: IMO we should remove all the inner JSON schemas and only generate schemas for the top-level credentials (i.e. the JWTs), but I don't want to remove any schemas that people may be relying on.

elenasantiago1 (approved):

PR #96, opened 2024-04-05
Fix: Add max value to UPRN to ensure it is generated as a Java Long

Joe-Edwards-GDS:
By default, the Java type generation maps the `integer` type to a Java `Integer`, however this is only a 32-bit integer which is not long enough for a UPRN.

Adding a maximum value above the max Java int persuades it to map to a `Long` instead.

### DIFF
--- a/v1/linkml-schemas/address.yaml
+++ b/v1/linkml-schemas/address.yaml
@@ -70,6 +70,7 @@ classes:
 slots:
   uprn:
     range: integer
+    minimum_value: 1
     maximum_value: 999999999999 # UPRNs can be up to 12 digits in length
     exact_mappings: adb:uprn
     description: The Unique Property Reference Number (UK and Northern Ireland Addresses Only)
--- a/v1/linkml-schemas/address.yaml
+++ b/v1/linkml-schemas/address.yaml
@@ -70,6 +70,7 @@ classes:
 slots:
   uprn:
     range: integer
+    maximum_value: 999999999999 # UPRNs can be up to 12 digits in length
     exact_mappings: adb:uprn
     description: The Unique Property Reference Number (UK and Northern Ireland Addresses Only)
                  <p>Exact mapping&#58; [adb:uprn](http://www.geoplace.co.uk/addressbase/schema/2.1/addressbase.xsd)</p>


Joe-Edwards-GDS:
An alternate approach is to just generate the Java schema with longs for _all_ number fields, but it feels a little unnecessary to have e.g. the GPG45 scores be long 🤷

danbudden-gds:
Couple of questions @Joe-Edwards-GDS .  

1. We know there are no negative integer UPRNs (the lowest is 1). Is it OK with you if we have a `minimum_value` as well (being `1`)? If so one of us can add this in. 

2. Is this approach OK when UPRNs have leading zeros? (I guess we need something in the java typing to ensure we don't lose leading zeros in the UPRN value when generating as Long?) For example the UPRN for [Bristol Central Library](https://en.wikipedia.org/wiki/Bristol_Central_Library) is 000000199356.

Joe-Edwards-GDS:
@danbudden-gds sorry I missed this comment - happy to add the `minimum_value`.

The leading zeroes question is an interesting one... If the leading zero is significant then we shouldn't use a numeric type at all (it'd need to be a string), which will have significant ramifications.

I vaguely recall that UPRNs shouldn't care about leading zeroes, but sometimes people format with them so they are a consistent length.

danbudden-gds:
> @danbudden-gds sorry I missed this comment - happy to add the `minimum_value`.
> 
> The leading zeroes question is an interesting one... If the leading zero is significant then we shouldn't use a numeric type at all (it'd need to be a string), which will have significant ramifications.
> 
> I vaguely recall that UPRNs shouldn't care about leading zeroes, but sometimes people format with them so they are a consistent length.

I think we should check if leading zeros are significant. Until we know for sure they are not, I'm tempted to be cautious, and assume they are, and push for this to be a string (as you suggest) to make sure we don't accept `199356` when it should be `000000199356` for [Bristol Central Library](https://en.wikipedia.org/wiki/Bristol_Central_Library). (And then we don't have min or max values but a min and max length.)

Can you elaborate on the ramifications you mention? If having string is an issue, I think I will revoke my proposal for a `minimum_value` as it introduces the risk of trimming leading zeros.

Joe-Edwards-GDS:
@danbudden-gds changing this to be a string would be a breaking change to the schema, so we'd need to coordinate across the Address CRI/IPV Core/SPOT etc. (and probably still need to support the old version as well).

At the moment it's a JavaScript `number`, which doesn't allow for leading zeroes anyway.

(i.e. the difference between `"uprn": 199356` and `"uprn": "000000199356"`)

danbudden-gds:
CDDO and Geoplace publish the UPRN standard. I found a [challenge PR](https://github.com/co-cddo/open-standards/issues/68#issuecomment-491245902) for the standard, and @Joe-Edwards-GDS found that they say 

> UPRNs and USRNs are integers (numbers) that can be up to 12 digits in length; they can therefore be less than 12 digits long and do not require leading zeros.

So we're good to go.

danbudden-gds (approved):
Based on the UPRN standard [as found here](https://github.com/co-cddo/open-standards/issues/68#issuecomment-491245902) we don't need leading zeros, so integer is good

PR #98, opened 2024-04-15
SPT-746:update Vocab schema to update Identity check type

chaitanguttarlapalli:
### What

When validating the credential in Spot we have failure for  Driving license Credential issuer (review-d), this is  because in the past driving license sent Identity check type  as IDENTIY_CHECK

### Why

Some old reused credentials still have IDENTITY_CHECK as Identity check type, we need to support both values IdentityCheck/IDENTTIY_CHECK values  for identity Check type.[SPT-746]

[SPT-746]: https://govukverify.atlassian.net/browse/SPT-746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v1/linkml-schemas/evidence.yaml
+++ b/v1/linkml-schemas/evidence.yaml
@@ -155,10 +155,10 @@ enums:
       money_laundering_regulations:
       physical_or_biometric_official:
   IdentityCheckType:
+    description: IDENTITY_CHECK is no longer used, historic value for backward compatibility reasons only use IdentityCheck instead
     permissible_values:
       IDENTITY_CHECK:
       IdentityCheck:
-
 slots:
   txn:
     description: A unique transaction identifier for this check, or part of a check, if any.
--- a/v1/linkml-schemas/evidence.yaml
+++ b/v1/linkml-schemas/evidence.yaml
@@ -16,7 +16,8 @@ classes:
   IdentityCheckClass:
     attributes:
       type:
-        equals_string: IdentityCheck
+        range: IdentityCheckType
+        description: Identity check type values allowed
     slots:
       - txn
       - strengthScore
@@ -153,6 +154,10 @@ enums:
       published: 
       money_laundering_regulations:
       physical_or_biometric_official:
+  IdentityCheckType:
+    permissible_values:
+      IDENTITY_CHECK:
+      IdentityCheck:
 
 slots:
   txn:
--- a/v1/linkml-schemas/evidence.yaml
+++ b/v1/linkml-schemas/evidence.yaml
@@ -155,9 +155,10 @@ enums:
       money_laundering_regulations:
       physical_or_biometric_official:
   IdentityCheckType:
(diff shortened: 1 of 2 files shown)


chris-dco (dismissed):

galund:
I guess this is an example where we need to be much more clear about how we do versioning.

I'd be tempted to say SPOT should not be validating using these schemas at all. Maybe it needs its own laxer schema that is derived in some consistent way from these ones. (eg by importing this one and making its own edits as a build time process).

Consider whether this schema could carry on being something stricter, to be followed by data producers, rather than a specification for data consumers working around strange old edge cases.

If we must include this here, I'd want to include some clear, publically-readable descriptions that explain which option is preferred, and which is just a backward-compatibility thing.

chrisclayson:
> I guess this is an example where we need to be much more clear about how we do versioning.
> 
> I'd be tempted to say SPOT should not be validating using these schemas at all. Maybe it needs its own laxer schema that is derived in some consistent way from these ones. (eg by importing this one and making its own edits as a build time process).
> 
> Consider whether this schema could carry on being something stricter, to be followed by data producers, rather than a specification for data consumers working around strange old edge cases.
> 
> If we must include this here, I'd want to include some clear, publically-readable descriptions that explain which option is preferred, and which is just a backward-compatibility thing.

@galund, what we've been doing with @cmiddi's grace, is bring v1 up to speed with "reality" which has meant feeding some "not ideal" things back into vocab 😢 

We should then introduce a new version which takes all the rubbish out and use that in an exercise in getting all CRIs "up to spec".

We did briefly discuss have a SPOT specific schema, I know I wasn't keen on that personally, but can't remember off the top of my head why we opted not to.

We could "downgrade" this to `v0` 😂 

What I am keen on, is us getting rid of the tech-debt of having the maintain our own schema in SPOT and to force all proposed schema changes to come via data-architecture and vocab and that the JSON schemas spat out here are then just imported into SPOT (hopefully as some kind of Dependabot managed dependency 🤞🏻  ... but ... baby steps).

v1/linkml-schemas/evidence.yaml
```diff
@@ -153,6 +154,10 @@ enums:
       published: 
       money_laundering_regulations:
       physical_or_biometric_official:
+  IdentityCheckType:
+    permissible_values:
+      IDENTITY_CHECK:
```
chrisclayson, line 159: We could add some metadata here, just for clarity:

```yaml
  IdentityCheckType:
    permissible_values:
      IDENTITY_CHECK:
        deprecated: true
        comments:
          - Historic value for backward compatibility reasons only.
          - Please use IdentityCheck instead.
      IdentityCheck:
```

chrisclayson: Not sure what the `deprecated: true` flag will have on the produced JSON schema though.

chrisclayson (approved):

PR #99, opened 2024-04-15
SPT-767 Made Mitigating validFrom datetime

chris-dco:
## What

Updated MitigatingCredentialClass.validFrom field to `datetime`.

## Why

CIMIT is required to produce `validFrom` in ISO time and date format. So the `validFrom` field needs to be changed to datetime.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -174,6 +174,7 @@ nav:
           - kbvQuality: v1/slots/kbvQuality.md
           - kbvResponseMode: v1/slots/kbvResponseMode.md
           - mitigatingCredential: v1/slots/mitigatingCredential.md
+          - mitigatingCredentialClass__validFrom: v1/slots/mitigatingCredentialClass__validFrom.md
           - mitigation: v1/slots/mitigation.md
           - name: v1/slots/name.md
           - nameParts: v1/slots/nameParts.md
--- a/scripts/test
+++ b/scripts/test
@@ -7,6 +7,7 @@ ajv validate -c ajv-formats -s $HERE/../v1/json-schemas/IdentityCheckCredential.
 ajv validate -c ajv-formats -s $HERE/../v1/json-schemas/IdentityCheckCredential.json -d $HERE/../v1/examples/driving-permit.json
 ajv validate -c ajv-formats -s $HERE/../v1/json-schemas/IdentityCheckCredentialJWT.json -d $HERE/../v1/examples/app-gdc-liveness-likeness-jwt.json
 ajv validate -c ajv-formats -s $HERE/../v1/json-schemas/IdentityCheckCredentialJWT.json -d $HERE/../v1/examples/identity-check-jwt.json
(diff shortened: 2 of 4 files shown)


chrisclayson (dismissed):
Wish your IDE would stop re-formatting things ... granted just getting rid of whitespace.

galund (changes requested):
Please at least make the whitespace changes in a different commit otherwise you break the git history for those lines. Right now it's not actually clear to me what this change is actually _doing_.

chris-dco:
> Please at least make the whitespace changes in a different commit otherwise you break the git history for those lines. Right now it's not actually clear to me what this change is actually _doing_.

@chrisclayson @galund If you click on the gear, Hide Whitespace and Apply and reload you can easily see the none whitespaces changes which have been made. Alternatively click on this link which will do it for you https://github.com/govuk-one-login/data-vocab/pull/99/files?diff=split&w=1.

github com_govuk-one-login_data-vocab_pull_99_files_diff=split w=1

chris-dco:
> Wish your IDE would stop re-formatting things ... granted just getting rid of whitespace.

@chrisclayson Sorry but I've had too many PR's rejected as I have inserted whitespaces and I understand our coding standards require us to not add whitespaces. It's also quite often hard enough to get anyone to review anything as it is.

galund:
> > Please at least make the whitespace changes in a different commit otherwise you break the git history for those lines. Right now it's not actually clear to me what this change is actually _doing_.
> 
> @chrisclayson @galund If you click on the gear, Hide Whitespace and Apply and reload you can easily see the none whitespaces changes which have been made. Alternatively click on this link which will do it for you https://github.com/govuk-one-login/data-vocab/pull/99/files?diff=split&w=1.
> 
> github com_govuk-one-login_data-vocab_pull_99_files_diff=split w=1

That's true but the point is that the Git history is useful. For YAML files especially (where trailing whitespace is *significant*) it's very much worth doing this as a separate commit please. Suggest you adjust your editor for YAML as automatically trimming whitespace can actually have unwanted effects!

chris-dco:
> That's true but the point is that the Git history is useful. For YAML files especially (where trailing whitespace is _significant_) it's very much worth doing this as a separate commit please. Suggest you adjust your editor for YAML as automatically trimming whitespace can actually have unwanted effects!

@galund Reverted whitespace changes.

galund (commented):
Can we get an example put into the `v1/examples` directory, and then get that validated in the tests?

Then we'd probably catch this kind of thing sooner - and we can easily check that the example matches the one in the RFC over here: https://github.com/govuk-one-login/architecture/blob/main/rfc/0041-CI-mitigation-VC.md

thanks!

chris-dco:
> Can we get an example put into the `v1/examples` directory, and then get that validated in the tests?
> 
> Then we'd probably catch this kind of thing sooner - and we can easily check that the example matches the one in the RFC over here: https://github.com/govuk-one-login/architecture/blob/main/rfc/0041-CI-mitigation-VC.md
> 
> thanks!

@galund @chrisclayson Added example and tests.

galund (approved):

PR #100, opened 2024-04-17
feat: Update InheritedIdentityJWT to contain an IdentityCheckCredential

Joe-Edwards-GDS:
The Inherited Identity JWT will always contain an `IdentityCheckCredential`.

The existing link to `VerifiableCredential` only includes the base VC fields, rather than a concrete implementation.


philf999 (approved):
Discussed this with @Joe-Edwards-GDS. Seems like a sensible change

PR #101, opened 2024-04-18
Generate Java Types

Joe-Edwards-GDS:
This is a rebase and update of https://github.com/govuk-one-login/data-vocab/pull/80

* Adds Java to the devcontainer configuration, so you can generate the Java types in it
* Adds a custom rules factory which generates types in the definitions sections of a json-schema
* Generates Java types for all credentials using a combined credentials schema
* Strips the `Class` suffix from generated types
* Fixes inheritance hierarchy by making `Person` a type of `CredentialSubject`
* Generates intermediate json-schema with `extends` to model inheritance


code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomSchemaRule.java
```diff
@@ -0,0 +1,49 @@
...
+import com.sun.codemodel.JClassContainer;
+import com.sun.codemodel.JType;
+import org.jsonschema2pojo.Schema;
+import org.jsonschema2pojo.rules.RuleFactory;
+import org.jsonschema2pojo.rules.SchemaRule;
+
+/**
+ * Custom schema rule that generates all classes contained within the '$defs' section of the json-schema.
+ * This may be removed once https://github.com/joelittlejohn/jsonschema2pojo/pull/1523 (or equivalent) is available.
+ */
```
Joe-Edwards-GDS, line 13: To elaborate a little more - generating from the `$defs` section means we can generate all the types from a single schema file, which avoids creating duplicate versions of the same classes.

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomObjectRule.java
```diff
@@ -0,0 +1,42 @@
...
+
+import static org.jsonschema2pojo.util.TypeUtil.resolveType;
+
+/**
+ * Custom object rule that includes interfaces for classes with a given suffix
+ */
+public class CustomObjectRule extends ObjectRule {
+    private static final Map<String, String> INTERFACE_MAPPING = Map.of(
+            "CredentialJWTClass", "uk.gov.di.model.CredentialJWTClass",
+            "CredentialClass", "uk.gov.di.model.CredentialClass"
```
Joe-Edwards-GDS, line 23: This is obviously not a scalable approach - it fixes the most fundamental problem of not having a shared interface for the different credential types (which is by far the most problematic), but:
- Those interfaces don't contain any content, so you still have to cast it to a specific type of credential to do anything with it
- It doesn't cover other relationships - e.g. it would be nice to capture the fact that `IdentityCheck extends PersonWithIdentity`, or `Passport extends Document`

Solving this fully is likely impossible with the current approach, as json-schema doesn't capture these relationships. 😞

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomObjectRule.java
```diff
@@ -0,0 +1,50 @@
...
+ * Custom object rule that:
+ * - strips the `Class` suffix from generated types (see main README)
+ * - adds interfaces for classes with a given suffix
+ */
+public class CustomObjectRule extends ObjectRule {
+    private static final String CLASS_SUFFIX = "Class";
+
+    private static final Map<String, String> INTERFACE_MAPPING = Map.of(
+            "CredentialJWTClass", "uk.gov.di.model.CredentialJWT",
+            "CredentialClass", "uk.gov.di.model.Credential"
```
Joe-Edwards-GDS: This is obviously not a scalable approach - it fixes the most fundamental problem of not having a shared interface for the different credential types (which is by far the most problematic), but:
- Those interfaces don't contain any content, so you still have to cast it to a specific type of credential to do anything with it
- It doesn't cover other relationships - e.g. it would be nice to capture the fact that `IdentityCheck extends PersonWithIdentity`, or `Passport extends Document`

Solving this fully is likely impossible with the current approach, as json-schema doesn't capture these relationships. 😞

pcornish (dismissed):

pcornish (approved):

PR #103, opened 2024-04-19
fix: Make person a valid credential subject

Joe-Edwards-GDS:
At the moment the inheritance hierarchy generated by the schema doesn't quite work for credential subjects.

For example:
- `IdentityCheckCredential` extends `VerifiableCredential`
- `IdentityCheckCredential.credentialSubject` is an `IdentityCheckSubject`
    - extends `PersonWithDocuments`
    - extends `PersonWithIdentity`
    - extends `Person`
- `VerifiableCredential.credentialSubject` is a `CredentialSubject`
- But `IdentityCheckSubject` is not a `CredentialSubject`, so the override for `credentialSubject` is not valid.

This can be fixed by making `Person` be a `CredentialSubject` (and this fixes it everywhere, as currently all `CredentialSubject`s are ultimately a `Person`). If we need other types of subject, we can do the same thing.

### DIFF
--- a/v1/linkml-schemas/common.yaml
+++ b/v1/linkml-schemas/common.yaml
@@ -42,6 +42,14 @@ classes:
         range: string
         description: The date in the form CCYY[-MM[-DD]]
 
+  CredentialSubjectClass:
+    description: Verifiable credential subjects use multiple inheritance; this base class is intentionally blank.
+    see_also:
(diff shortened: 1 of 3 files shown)


pcornish (approved):
LGTM - this also fixes the generation of a single-file JSON schema.

Would be good to have @govuk-one-login/data-architects review too.

Joe-Edwards-GDS:
Superseded by #101

PR #106, opened 2024-04-26
refactor(ts-types): TypeScript generator uses combined JSON Schema.

pcornish:
Uses a combined JSON Schema document as the input for TypeScript type generation.

# Background
- The Java type generator uses a combined JSON Schema, as of #101 
- Historically, the TypeScript generator used separate schema files, which led to duplication of types

# What changed
- The TypeScript generator now uses a combined schema file as its input, to align it to the Java generator
- This has allowed further file structure simplifications, following #104 

# Testing it
- Follow the prerequisites in `code-generators/typescript-types/README.md`
- Run `scripts/generate_ts_types.sh`, which
  - produces a combined schema file under `code-generators/typescript-types/schemas`
  - generate types under  `code-generators/typescript-types/dist`
- Test with the example project (`examples/typescript-node`):
  - Replace the dependency in `package.json` with a local dependency: `npm install ../../code-generators/typescript-types/dist`
  - Run the example project (see its readme), which consumes it without any type errors

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -56,7 +56,8 @@ jobs:
         run: scripts/test
 
       - name: Generate TypeScript types
-        run: npm run generate
+        working-directory: scripts
+        run: ./generate_ts_types.sh
 
       - name: Generate Java types
         working-directory: scripts
--- a/.gitignore
+++ b/.gitignore
@@ -27,7 +27,3 @@ site
 .DS_Store
 
 Staticfile.auth
-
-# Generated TypeScript types
-gen/
-code-generators/typescript-types/dist/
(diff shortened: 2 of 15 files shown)


Joe-Edwards-GDS (commented):
I think it makes sense to generate the Java and TS types in as similar a way as possible, and this looks broadly sensible! :+1:

Some observations:
- It still seems to generate multiple copies of some enum types (even though they are being generated from a shared schema), but I haven't tried to diagnose this. It's not a major issue because they're equivalent in typescript anyway :shrug:
- TS does get upset about the inheritance in one place - `ISODateClass` is a `StructuredAttributeClass`, but they disagree on whether `value` is optional. Hopefully easy enough to fix in the schema
- It would be nice if we could get rid of the `Class` suffixes!

pcornish:
> ...
> * It would be nice if we could get rid of the `Class` suffixes!

Yes, I agree 👍 

I've raised #112 to cover this.

PR #111, opened 2024-05-24
SPT-808: Schema only typescript distribution package

chrisclayson:


code-generators/npm-schema-only-package/src/package-template.json
```diff
@@ -0,0 +1,12 @@
+{
```
chris-dco: In the top level `package.json` there is a `scripts` section `copy-pkg-misc` which copies the `package-template.json` and `README.md` to `code-generators/typescript-types/dist`. It would be worth creating two new script sections one to copy the `typescript-types` files (e.g. `copy-pkg-typescript-types`) and another to copy the `npm-schema-only` package (e.g. `copy-npm-schema-only`), the update the `copy-pkg-misc` to run them both e.g. `copy-pkg-typescript-types && copy-npm-schema-only`.

code-generators/npm-schema-only-package/src/index.mjs
```diff
@@ -0,0 +1,60 @@
...
+const genDir = args.length >= 2 ? args[1] : path.join(process.cwd(), "gen");
+if (!fs.existsSync(genDir)) {
+    await mkdir(genDir);
+    await mkdir(path.join(genDir, 'schemas'));
+}
+
+generate(schemaDir, genDir)
+    .catch((cause) => console.error(cause));
+
+await writeFile(path.join(genDir, 'package.json'), JSON.stringify({
```
chris-dco: See comment in https://github.com/govuk-one-login/data-vocab/pull/111#discussion_r1618371937 which may be a better alternative.

chrisclayson: Weirdly, I did it this way to avoid have the external scripts and template, thought it was neater to wrap up into the generation script 🤔

code-generators/npm-schema-only-package/src/package-template.json
pcornish: As part of #106, the top level scripts have been moved down into https://github.com/govuk-one-login/data-vocab/blob/main/code-generators/typescript-types/package.json

This is more cohesive as those steps applied to the sub-project. Their presence in the top-level `package.json` was a vestige of an older structure for the type generators in this repo.

sam-leen-gds (approved):

PR #113, opened 2024-06-27
Adding example using ajv and vocab schema

richa-misra:


chrisclayson (approved):
Thanks for doing this ... good example of how to use this (and, coincidentally, pretty much SPOT will doing with it 😄 )

PR #117, opened 2024-07-11
Added decisionScore to IdentityCheck

GDS-Dominic:
I have noticed the Fraud CRI VC Issued event contains this `decisionScore` attribute which isn't present here. I'm not familiar with this enough to know if this should be added here but raising a PR to bring attention to it.

It would also be good if someone can add in a minimum and maximum value and review the description.


GDS-Dominic:
Not needed

PR #123, opened 2024-10-28
Add addressRegion and poBoxNumber to address schema

danbudden-gds:
This PR addresses (heh) [DIAR-2048](https://govukverify.atlassian.net/browse/DIAR-2048) and [DIAR-176](https://govukverify.atlassian.net/browse/DIAR-176).

While the net result is two new optional attributes, there have been some under the hood changes to classes.
- new `PostalAddressGeneralClass` as an abstract class with attributes common to the UK and other countries (we put `poBoxNumber` here because the logical attribute is in both Ordnance Survey AddressBase and schema.org)
- new `PostalAddressUKClass` which inherits the general class and adds attributes only used in UK (e.g. `uprn`)
- new `PostalAddressInternationalClass` which inherits the general class and adds attributes not typically used in UK (nor found in Ordnance Survey AddressBase) and which are primarily driven by the need to support non-UK addresses (for now, only `addressRegion` but more may be added in future)
- `PostalAddressClass` now inherits its attributes as mix-ins from both the UK and non-UK classes named above, to result in the all-purpose class for postal address

image

The rationale for doing the modelling this way is to preserve the UK-specific examples and to add new non-UK examples in a separate dedicated place in the UI, while maintaining the `PostalAddressClass` as the reference for other classes and schema to get their address stuff (mainly `person.yaml`).

For now, everything still runs off `PostalAddressClass` as before. If later we identify that we could or should reference just the UK or non-UK schema in certain places in the vocabulary, the classes are now there to support that. 

Local deployment looks sensible. Added `addressRegion` to the `data-vocab/v1/examples/core-cri-authz-request.json` test file and tests pass locally. 

_Screengrab from local deployment_

image

[DIAR-2048]: https://govukverify.atlassian.net/browse/DIAR-2048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DIAR-176]: https://govukverify.atlassian.net/browse/DIAR-176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Joe-Edwards-GDS:
It's a clever approach, but worth observing that the JSON schemas (and therefore the TS and Java types) only get the combined `PostalAddressClass`. This means that although the UK/International distinction is represented in the vocab site (and linkml), consumers won't currently be able to make use of it - limiting its value.

I haven't gone into great detail, but I suspect this may not be straightforward to resolve, as mixins map very differently across schemas and languages!

danbudden-gds:
> It's a clever approach, but worth observing that the JSON schemas (and therefore the TS and Java types) only get the combined `PostalAddressClass`. This means that although the UK/International distinction is represented in the vocab site (and linkml), consumers won't currently be able to make use of it - limiting its value.
> 
> I haven't gone into great detail, but I suspect this may not be straightforward to resolve, as mixins map very differently across schemas and languages!

Yes, agreed that is a limitation. I felt constrained by `person.yaml` wanting to use a single class as a [range](https://github.com/govuk-one-login/data-vocab/blob/aadba214242d7dc5556addec390197ee4f2d72cc/v1/linkml-schemas/person.yaml#L70) for its address slot hence joining everything back into `PostalAddressClass`. I felt if I started changing person.yaml, it's picking at the thread too much. 

The benefit is mainly just documentation at the moment, so we can have pages like the below, where I can have example values for `addressRegion` populated. (Because I am _not_ keen on specifying permitted values for it...)

image

image

Perhaps I should expose the UK and International classes into json and TS schema - if that's useful and required? I don't have sight of how they could be used though.

Joe-Edwards-GDS (approved):
Happy to go with this, given the caveat above.

(I'd also be happy with a single address class with optional fields)

Caitlin-cooling (approved):

CoralSmith (approved):

chrisclayson (approved):
I can't see an issue with this ... proof will be what the JSON schemas look like when they're generated.

PR #125, opened 2024-11-26
Pyic 7650 add applicable_authoritative_source fraudCheck type

DanCorderIPV:
Add `country_data` as a possible value for FraudCheckType

I ran the build and test scripts successfully but they didn't appear to update anything that gets committed.

### DIFF
--- a/v1/linkml-schemas/evidence.yaml
+++ b/v1/linkml-schemas/evidence.yaml
@@ -141,6 +141,7 @@ enums:
       multiple_choice:
   FraudCheckType:
     permissible_values:
+      country_data:
       identity_theft_check:
       impersonation_risk_check:
       mortality_check:
--- a/v1/linkml-schemas/evidence.yaml
+++ b/v1/linkml-schemas/evidence.yaml
@@ -141,8 +141,8 @@ enums:
       multiple_choice:
   FraudCheckType:
     permissible_values:
-      country_data:
-        description: Denotes whether the third parties used for fraud checks are an authoritative source for data from that country. This means that if we fail, we cannot do an ID Fraud check and provide an identityFraudScore for that user.
+      applicable_authoritative_source:
+        description: Denotes whether the third parties used for fraud checks are an authoritative source for the current user. This means that if we fail, we cannot do an ID Fraud check and provide an identityFraudScore for that user.
       identity_theft_check:
       impersonation_risk_check:
       mortality_check:
--- a/v1/linkml-schemas/evidence.yaml
+++ b/v1/linkml-schemas/evidence.yaml
@@ -142,6 +142,7 @@ enums:
   FraudCheckType:
     permissible_values:
       country_data:
+        description: Denotes if a user has an international address
       identity_theft_check:
       impersonation_risk_check:
       mortality_check:
--- a/v1/linkml-schemas/evidence.yaml
+++ b/v1/linkml-schemas/evidence.yaml
@@ -141,10 +141,10 @@ enums:
       multiple_choice:
   FraudCheckType:
     permissible_values:
-      mortality_check:
       identity_theft_check:
-      synthetic_identity_check:
       impersonation_risk_check:
+      mortality_check:
+      synthetic_identity_check:
   DataCheckType:
     permissible_values:
       record_check:
--- a/v1/linkml-schemas/evidence.yaml
+++ b/v1/linkml-schemas/evidence.yaml
@@ -142,7 +142,7 @@ enums:
   FraudCheckType:
     permissible_values:
       country_data:
-        description: Denotes if a user has an international address
+        description: Denotes whether the third parties used for fraud checks are an authoritative source for data from that country. This means that if we fail, we cannot do an ID Fraud check and provide an identityFraudScore for that user.
       identity_theft_check:
       impersonation_risk_check:
       mortality_check:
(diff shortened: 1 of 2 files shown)


blakeyp (dismissed):

v1/linkml-schemas/evidence.yaml
```diff
@@ -141,10 +141,12 @@ enums:
       multiple_choice:
   FraudCheckType:
     permissible_values:
-      mortality_check:
+      country_data:
+        description: Denotes if a user has an international address
```
CoralSmith: Can we be a little clearer about what this check does?

This check denotes whether the third parties used for fraud checks are an authoritative source for data from that country. 

This means if we fail we cannot do an ID Fraud check and provide an `identityFraudScore` for that user.

GDS-Dominic (dismissed):
Feels better with the description, I am approving but I would like @danbudden-gds  to give a once over as he is more involved with the identity side of things

v1/linkml-schemas/evidence.yaml
danbudden-gds: > whether the third parties used for fraud checks are an authoritative source for data from that country.

Got a couple of Qs so I can come to an opinion on the new `permissible_value`. 

If these Qs are answered in business requirements somewhere, apologies, ~~if you could link me I'll go away and read up~~ [edit: @MikeCollingwood has given me the [ticket](https://govukverify.atlassian.net/browse/PYIC-7650) so I'll read up first then chase with any residual questions]. 

1. What piece of data are we submitting for a check. 
2. What kind of endpoint tells us the answer - is it polling authoritative sources themselves, or some config or reference that we own?
3. What is the logic of the check? 

Is it something like "user is submitting UK-issued documents, so we're checking that (e.g.) experian is a valid option for KBV against our reference source config?" 

4. What would a VC look like where `fraudCheck: country_data` appears (roughly, don't need full example) - is/are there other attributes that denote a success or fail? 
5. Presumably we wouldn't have both `fraudCheck: country_data` and `fraudCheck: identity_theft_check` in the same VC, so I'm just trying to get clued up on the process. The VC with `fraudCheck: country_data` is done prior to, or same time as, other fraud checks? 

I'm thinking that `country_data` is not as self-descriptive as some of the other `permissible_value` items if somebody doesn't have the above context, but I'm struggling to suggest an alternative without knowing the context. It might be as simple as `country_data_check`.

danbudden-gds: I think we may want to slightly edit [line 238](https://github.com/govuk-one-login/data-vocab/blob/92dd976b11a6d30aa23c0804873d35792c4f4851/v1/linkml-schemas/evidence.yaml#L238C119-L238C270) as well, because the enums of this check no longer exclusively mean a GPG45 thing? But I'm not 100% sure on that - depends on the above comment.

danbudden-gds: @MikeCollingwood has given me the [ticket](https://govukverify.atlassian.net/browse/PYIC-7650) so I'll read up first then chase with any residual questions

CoralSmith:
I'll come in and say I'll answer your questions directly @danbudden-gds

v1/linkml-schemas/evidence.yaml
CoralSmith: I've tried to move the conversation out of the PR and answer some of the questions to the above. We can keep the conversation going on the ticket if that's our preferred non-github focal point?

danbudden-gds:
Thanks @MikeCollingwood and @CoralSmith  I'm up to speed now. 

Now I know what's going on, I do feel that we can improve on `country_data` to make the VC more self-describing. 

How about, instead of `country_data` we use `authoritative_source_available` or `authoritative_source_availability`?
Giving: 

```
failedCheckDetails: [{
  {
    "checkMethod": "data",
    "fraudCheck": "authoritative_source_available"
  },
}]
```

As a bonus this would help to - as per the requirements:

> set us up nicely for a future use case for “fraud resilience” where we plan to support M1C in the case of Experian fraud being down.

danbudden-gds:
Or - if IPV or SPOT needs to distinguish between the two situations rather than use one value to cover both: 
1. At the time we're writing this VC we don't have any authoritative source we can use for a user presenting international address and/or documents
2. We do have an authoritative source we could use but there's a problem with getting to it right now

Then, for now (1) how about

`applicable_authoritative_source`

which can be used for other situations an AS might not apply, if we come across those, not just country

and for (2) later in another PR I expect:

`available_authoritative_source` 

which is more specific to resilience

DanCorderIPV:
@danbudden-gds your suggestion makes sense to me, but to do it will we need to update the relevant ADR: https://github.com/govuk-one-login/architecture/blob/main/adr/0196-fraud-checking-non-uk-residents.md ?

CoralSmith:
> Or - if IPV or SPOT needs to distinguish between the two situations rather than use one value to cover both: [...]

Lets go with that suggestion. For now this PR will add in `applicable_authoritative_source` & I'll go update the ADR.

DanCorderIPV:
I've updated the new value and tweaked the comment to not mention country data as presumably now this value could be used for more reasons than just that in the future.

blakeyp:
We've updated this per the updated ADR196, @danbudden-gds could we get a re-review please?

CoralSmith (approved):

CoralSmith:
I've approved it, the description is accurate and the value is what we've agreed, and is captured in the updated ADR that was approved by @danbudden-gds. 
I think this means we're good to merge this in.

PR #128, opened 2024-12-11
BAU: Remove `v` prefix when publishing packages

chrisclayson:
When publishing packages, ensure the `v` prefix is removed from the version name (the `v` should be retained while creating the GitHub release).

The inclusion of the `v` is proving troublesome with the Maven repository, NPM is clever enough to remove `v` but the Gradle publish action just uses the version verbatim and downstream this causes Dependabot to do strange things.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -93,7 +93,8 @@ jobs:
       - name: Calculate release name
         id: release-name
         run: |
-          echo "release_version=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_OUTPUT
+          TAG="${GITHUB_REF/refs\/tags\//}"
+          echo "release_version=${TAG#v}" >> $GITHUB_OUTPUT
 
       - name: Download combined schema
         uses: actions/download-artifact@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -93,7 +93,8 @@ jobs:
       - name: Calculate release name
         id: release-name
         run: |
-          echo "release_version=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_OUTPUT
+          TAG="${GITHUB_REF/refs\/tags\//}"
+          echo "release_version=${TAG#v}" >> $GITHUB_OUTPUT
```
Joe-Edwards-GDS, line 97: (Presumably you could also do `release_version=${GITHUB_REF/refs\/tags\/v/}`, but this is clearer :) )

Joe-Edwards-GDS (approved):

CoralSmith (approved):

.github/workflows/publish.yml
chrisclayson: My concern with this was that it assumes the `v` is always present, doing in two steps ensures the `/refs/tags/` is always removed ... could have made the `v` and optional capture I suppose 🤦🏻 I'll leave as is for the mo, it can be iterated on.

PR #131, opened 2025-02-03
PYIC-7917: add available_authoritative_source fraudCheck type

shivanshuit914:
add `available_authoritative_source`  fraudCheck type

### DIFF
--- a/v1/linkml-schemas/evidence.yaml
+++ b/v1/linkml-schemas/evidence.yaml
@@ -143,6 +143,8 @@ enums:
     permissible_values:
       applicable_authoritative_source:
         description: Denotes whether the third parties used for fraud checks are an authoritative source for the current user. This means that if we fail, we cannot do an ID Fraud check and provide an identityFraudScore for that user.
+      available_authoritative_source:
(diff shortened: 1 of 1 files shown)


v1/linkml-schemas/evidence.yaml
```diff
@@ -143,6 +143,8 @@ enums:
     permissible_values:
       applicable_authoritative_source:
         description: Denotes whether the third parties used for fraud checks are an authoritative source for the current user. This means that if we fail, we cannot do an ID Fraud check and provide an identityFraudScore for that user.
+      available_authoritative_source:
+        description: Denotes whether the third parties used for fraud checks are unavailable (for an example Experian). This means that if we fail, we can do an ID check and provide an identityScore for medium confidence user.
```
DanCorderIPV: We're a little sensitive about fraud checks, so maybe better to keep to the form of the comment above?
`Denotes whether the third parties used for fraud checks are available. This means that if we fail, we cannot do an ID Fraud check and provide an identityFraudScore for that user.`

shivanshuit914: Thanks Dan, Updated it :)

DanCorderIPV: Looks good to me :)

DanCorderIPV (approved):

PR #133, opened 2025-03-03
SPT-1296: Update data-vocab to have a GSON annotated model and fixes to builders

chris-dco:
## What
This change to the data-vocab generators resolves the following issues. Please see the Why section to understand why these are an issue.
* Adds a new sub-project `gson-annotated` which includes the GSON annotations
* Removes duplicate fields between inherited types
* Removes use of generics in Builders

As this is a complex subject please see the section Understanding this Pull Request to get an explanation for the changes.

## Why
Some components of OneLogin use GSON to serialise and deserialise JSON in Java. We looked into using the `uk.gov.di.model.di-data-model` package with GSON, however we found that we needed to:
* Implement a large number of work arounds in order to serialise/deserialise
* Add a `ExclusionStrategies` as fields such as `credentialSubject` were duplicate, such as the following. Unlike Jackson, GSON will not allow duplicate fields to exist and will throw an exception:
  * `VerifiableCredential` defined a field `credentialSubject:CredentialSubject`
  * `VerifiableIdentityCredential` which extends `VerifiableCredential` also defined a field `credentialSubject:PersonWithIdentity`
* The Builders generated by `uk.gov.di.model.di-data-model` used generics heavily, required a lot of type casting, as well as liberal use of `@SuppressWarnings`.

### GSON Serialiser workarounds
In order to create the GSON builder a lot of `TypeAdapters` and `ExclusionStrategies` needed to be used in order to serialise and deserialise data, for example:

<details>
<summary>Work around code using `uk.gov.di.model.di-data-model` with GSON</summary>

```java
    private static Gson getGson() {
        if (gson == null) {
            gson =
                    new GsonBuilder()
                            .setPrettyPrinting()
                            .registerTypeAdapter(PublicKey.class, new PublicKeyAdapter())
                            .registerTypeAdapter(
                                    IdentityVectorOfTrust.class,
                                    new EnumValueAdapter<>(
                                            IdentityVectorOfTrust::fromValue,
                                            IdentityVectorOfTrust::value))
                            .registerTypeAdapter(
                                    NamePartType.class,
                                    new EnumValueAdapter<>(
                                            NamePartType::fromValue, NamePartType::value))
                            .registerTypeAdapter(
                                    VerifiableCredentialType.class,
                                    new EnumValueAdapter<>(
                                            VerifiableCredentialType::fromValue,
                                            VerifiableCredentialType::value))
                            .setExclusionStrategies(new VerifiableCredentialExclusionStrategy())
                            .create();
        }
        return gson;
    }

    static class EnumValueAdapter<T extends Enum<T>> extends TypeAdapter<T> {
        private final Function<String, T> fromValue;
        private final Function<T, String> toValue;

        public EnumValueAdapter(Function<String, T> fromValue, Function<T, String> toValue) {
            this.fromValue = fromValue;
            this.toValue = toValue;
        }

        @Override
        public void write(JsonWriter out, T value) throws IOException {
            out.value(value != null ? toValue.apply(value) : null);
        }

        @Override
        public T read(JsonReader in) throws IOException {
            var token = in.peek();
            if (token.equals(JsonToken.STRING)) {
                return fromValue.apply(in.nextString());
            }
            throw new MalformedJsonException("Should be a string");
        }
    }
```
</details>

### Required type casting in builders
When using builders such as `VerifiableIdentityCredentialBuilder` you have to be very careful what order you call the `.with` methods. The `VerifiableIdentityCredentialBuilder` inherits from the generic builder `VerifiableCredentialBuilderBase`. Methods defined by `VerifiableIdentityCredentialBuilder` will return a `VerifiableIdentityCredentialBuilder`, however methods declared by `VerifiableCredentialBuilderBase` will return a `VerifiableCredentialBuilderBase`. So in this instance you should call:
1. all the methods within `VerifiableIdentityCredentialBuilder` followed by the methods within `VerifiableCredentialBuilderBase`.
2. When you call the `build` method remember your now calling the `build` method on `VerifiableCredentialBuilderBase` and not `VerifiableIdentityCredentialBuilder`, so you will need to cast the result from the build method to `VerifiableIdentityCredential`.

### Warning: Unchecked call to 'withNameParts(List<NameParts>)' as a member of raw type 'uk.gov.di.model.NameBuilderBase'

For code such as the following you would very often hit a warning similar to `Unchecked call to 'withNameParts(List<NameParts>)' as a member of raw type 'uk.gov.di.model.NameBuilderBase'` which can only be resolved by suppressing it was `@SuppressWarning("unchecked")`.

```java
Name.builder()
    .withValidFrom(name.getValidFrom())
    .withValidUntil(name.getValidUntil())
    .withNameParts(...)
    .build();
```

This is due to the heavy use of generics generated by the builders. For example, the NameBuilder is defined as follows with a heavy use of generics. In most cases the `NameBuilder` will only ever need to create type of `Name` however the default behaviour of the generator is to use generics.

```java
    public static class NameBuilder extends NameBuilderBase<Name> {
        public NameBuilder() {
        }
    }

    public abstract static class NameBuilderBase<T extends Name> {
        protected T instance;
```

This gets even more complex in scenarios such as cases such as `VerifiableIdentityCredential` which inherits from `VerifiableCredential` resulting in the following builder code being generated.

<details>

<summary>Sample code automatically generate for `uk.gov.di.model.di-data-model`</summary>

```java

    public static class VerifiableIdentityCredentialBuilder extends VerifiableIdentityCredentialBuilderBase<VerifiableIdentityCredential> {
        public VerifiableIdentityCredentialBuilder() {
        }
    }

    public abstract static class VerifiableIdentityCredentialBuilderBase<T extends VerifiableIdentityCredential> extends VerifiableCredential.VerifiableCredentialBuilderBase<T> {
        public VerifiableIdentityCredentialBuilderBase() {
            if (this.getClass().equals(VerifiableIdentityCredentialBuilder.class)) {
                this.instance = new VerifiableIdentityCredential();
            }
        }

    public static class VerifiableCredentialBuilder extends VerifiableCredentialBuilderBase<VerifiableCredential> {
        public VerifiableCredentialBuilder() {
        }
    }

    public abstract static class VerifiableCredentialBuilderBase<T extends VerifiableCredential> {
        protected T instance;
```
</details>

## Understanding this Pull Request

The data-vocab project uses the library `org.jsonschema2pojo` to convert between JSON Schema and Plain Old Java Objects (POJO). The library uses the Sun Code Model to model classes, fields, methods and code which are then complied to form the following libraries:
* uk.gov.di.model.di-data-model - Provides Java POJO and Builders which models the data-vocab. 
* uk.gov.di.model.di-data-model-jackson - Provides Jackson annotated POJO which models the data-vocab.

At the heart we provide a `CustomRuleFactory` which enables us to hook into the various lifecycle events which occur whilst converting between the JSON Schema and Java. Originally we had the following classes:
* CustomRuleFactory - Replaced the default `SchemaRule` with the `CustomObjectRule` and the default `ObjectRule` with the `CustomObjectRule` classes.
* CustomObjectRule - Many of the classes defined within the data-vocab JSON Schema are postfixed with `Class`. This rule renames these classes removing the `Class` postfix and keeping the rest of the name.
* CustomSchemaRule - Schema rule to ensure that 'all' classes contained within the `$defs` section of the JSON Schema are created.

As we mentioned earlier this PR make the following changes:
* Adds a new sub-project `gson-annotated` which includes the GSON annotations
* Removes duplicate fields between inherited types
* Removes use of generics in Builders

### Adding the gson-annotated sub-project

The gson-annotated sub-project is fairly self explanatory and it based upon the jackson-annotated sub-project. The only differences are:
* The annotation style within the `build.gradle` is set to `gson`.
* The `generateBuilders` and `useInnerClassBuilders` org.jsonschema2pojo options are enabled in the build.gradle.
* A number of additional tests have been added to ensure that serialisation, deserialisation and builders are working as expected without work arounds.

### Removes duplicate fields between inherited types

Removing inherited fields is implemented within the `CustomPropertyRule` which extends from the default `PropertyRule`. This rule is called every time a new field is added to a class. The overridden behaviour will test to see if this class extends from another class and whether that field already exists. If the field already exists in the parent class then it will not be added. This is done by not calling the `super.apply` otherwise the field, getters and setters will all be added.


code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomNameHelper.java
Joe-Edwards-GDS, line 1: Not sure why I didn't spot this before - this is obviously how you should fix the names.

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomBuilderRule.java
Joe-Edwards-GDS, line 1: This is pretty complicated - I'd probably err on the side of over-commenting, because it's going to be really hard for someone to understand if they come back to this in future.

It looks like a lot of this is just copy-pasted from the original implementation, in which case I'd suggest linking to the original, and clearly marking where you've changed something. Ideally keeping things as close to the original as possible where you haven't. That way if the library updates, it's easier to copy across your changes.

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomBuilderRule.java
```diff
@@ -0,0 +1,106 @@
...
+import com.sun.codemodel.JVar;
+import org.jsonschema2pojo.Schema;
+import org.jsonschema2pojo.rules.Rule;
+import org.jsonschema2pojo.rules.RuleFactory;
+import org.jsonschema2pojo.util.ReflectionHelper;
+
+import java.util.Objects;
+
+/**
+ * Custom implementation of the BuilderRule which generates a Builder without needing generics.
```
Joe-Edwards-GDS: Can you give an example of before/after?

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomObjectRule.java
```diff
@@ -1,39 +1,173 @@
...
+import java.util.Iterator;
 
-import static org.jsonschema2pojo.util.TypeUtil.resolveType;
+import static uk.gov.di.codegen.CustomUtils.toStream;
 
 /**
  * Custom object rule that:
  * - strips the `Class` suffix from generated types (see main README)
  * - adds interfaces for classes with a given suffix
  */
```
Joe-Edwards-GDS, line 30: This is no longer accurate. As above - it's probably worth trying to explain a bit more what's going on here (perhaps with examples of the generated code being targeted), because AST manipulation is really tricky to follow.

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomObjectRule.java
```diff
@@ -1,39 +1,173 @@
...
+                override = typeParam._extends();
+            }
+
+            narrowedObject = narrowedObject._extends(superType.narrow(override));
+        }
+    }
+
+    private void makePrivateMembersProtected(JDefinedClass jClass) {
+        for (var jField : jClass.fields().values()) {
+            if (jField.mods().getValue() == 4) {
```
Joe-Edwards-GDS: `== JMod.PRIVATE`?

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomObjectRule.java
```diff
@@ -1,39 +1,173 @@
...
+            JType override = foundOverrides.get(typeParam.name());
+            if (override == null) {
+                override = typeParam._extends();
+            }
+
+            narrowedObject = narrowedObject._extends(superType.narrow(override));
+        }
+    }
+
+    private void makePrivateMembersProtected(JDefinedClass jClass) {
```
Joe-Edwards-GDS, line 152: Why are we making everything protected?

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomSchemaRule.java
```diff
@@ -26,24 +31,56 @@ public JType apply(String nodeName, JsonNode schemaNode, JsonNode parent, JClass
         // Process schema as normal
         var type = super.apply(nodeName, schemaNode, parent, generatableType, schema);
 
-        // Get definitions
-        final var definitions = schemaNode.at(DEFINITIONS_PATH).fields();
+        if (type instanceof JDefinedClass jDefinedClass && generatableType instanceof JDefinedClass jGeneratableType && isGenericField(nodeName, schema)) {
```
Joe-Edwards-GDS, line 50: What's this doing?

Joe-Edwards-GDS (commented):
Impressive work - this is very fiddly stuff!

It's a bit hard to review the detail without understanding the full intent behind it, but the proof is in the pudding with this sort of thing.

Looks like this changes the output for POJO and Jackson versions of the lib as well - which means changes required in IPV Core. Do you expect any of these to be significant, is there anything we should be worried about? At a glance it's only the generification of `VerifiableCredential` that causes any changes.

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomPropertyRule.java
```diff
@@ -0,0 +1,60 @@
...
+        super(ruleFactory);
+        this.ruleFactory = ruleFactory;
+    }
+
+    @Override
+    public JDefinedClass apply(String nodeName, JsonNode node, JsonNode parent, JDefinedClass jclass, Schema schema) {
+        String propertyName = this.ruleFactory.getNameHelper().getPropertyName(nodeName, node);
+
+        JClass jParentClass = jclass._extends();
+        if (jParentClass instanceof JDefinedClass jParentDefinedClass) {
```
michaelabenyohai: I'm not understanding this for some reason, please can you help where I am going wrong?

In my head our example is (from how things work currently):

- `VerifiableCredential` class
  - has `CredentialSubject credentialSubject` property
- `VerifiableIdentityCredential` class
  - extends `VerifiableCredential` class
  - has `PersonWithIdentity credentialSubject` property
  - `PersonWithIdentity` class extends `CredentialSubject` class

Our aim is to ensure the `PersonWithIdentity credentialSubject` property is not added to the `VerifiableIdentityCredential` class, I think?

In this code:
```java
JClass jParentClass = jclass._extends();
if (jParentClass instanceof JDefinedClass jParentDefinedClass) {
  ...
}
```

- jDefinedClass is `VerifiableIdentityCredential`
- jParentClass is `VerifiableCredential`

But if that is the case then I would expect `jParentClass instanceof JDefinedClass` to be `false`?

But I assume it must be true as we want it to reach the code in the if statement. Where am I going wrong?

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomPropertyRule.java
```diff
@@ -0,0 +1,60 @@
...
+        if (parentType.equals(childType)) {
+            return true;
+        }
+        if (childType instanceof JDefinedClass jDefinedChildClass && parentType instanceof JDefinedClass jDefinedParentClass) {
+            return isInheritedOrEqual(parentType, jDefinedChildClass._extends());
+        }
+        return false;
+    }
+
+    private JType getPropertyType(String nodeName, JsonNode node, JsonNode parent, JDefinedClass jclass, Schema schema) {
```
michaelabenyohai: Worth adding a comment that this is basically copied from the original `PropertyRule.java` rule

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomPropertyRule.java
Joe-Edwards-GDS: You're mixing classes and variables that represent classes! `JDefinedClass` is a code model type representing a class declaration, not a specific class itself.

`jParentClass instanceof JDefinedClass` is asking 'is `jParentClass` a defined class' (i.e. a class that we've defined, as opposed to, say, `Object`), which will be true for any type with a defined parent (e.g. `VerifiableIdentityCredential`).

Meta-programming is hard 😅

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomBuilderRule.java
chris-dco: I've added some comments on each of the files which hopefully should explain things going forwards.

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomBuilderRule.java
chris-dco: I've added some comments on each of the files which hopefully should explain things going forwards.

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomObjectRule.java
chris-dco: Changed to use `JMod.PRIVATE` and updated other instances.

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomObjectRule.java
chris-dco: Comments added including reasons why this was added.

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomSchemaRule.java
chris-dco: Added comments. Let me know if you need any other information.

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomPropertyRule.java
chris-dco: Comments added.

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomPropertyRule.java
chris-dco: Comments added.

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomObjectRule.java
Joe-Edwards-GDS: Thanks, I think 'the Builder needs to directly write to the fields of the parent class' was what I was missing.

I don't _love_ this - it breaks encapsulation and relies on you manually removing shadowed properties, but I can see that we don't really want to get deep into the weeds of `PropertyRule` to fix it properly :(

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomBuilderRule.java
Joe-Edwards-GDS: Thanks - this might be silly, but is there a reason we've reformatted the original code? E.g. the original has a bunch of helpful comments in it https://github.com/joelittlejohn/jsonschema2pojo/blob/master/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/rules/BuilderRule.java

It's a bit on-the-nose, but in the past I've done stuff like this to make it super clear:
```
// This class is an exact copy of [url to exact revision] with some modifications to do blah
public class SomeClass {
  private void someMethod() {
     ...
  }

  private void someOtherMethod() {
    var foo = "something";

    // #########################
    // Modified code starts here
    // #########################
    foo = "something else";
    // #########################
    // Modified code ends here
    // #########################

    return foo;
  }
}
```

I don't really own this codebase though, and what you've got is fine.

Joe-Edwards-GDS (approved):
Thanks for clarifying - left a couple more comments but don't want to hold you up any further (and I shouldn't be a blocker anyway!) so happy to go with this and refine later if need be.

code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomPropertyRule.java
michaelabenyohai: Thank you! I was being an idiot 🤦‍♀️

michaelabenyohai: Thank you! I was being an idiot 🤦‍♀️

michaelabenyohai (approved):

PR #145, opened 2025-09-09
Remove LinkML as a cached dependency and use live

huwd:
This is a proposal, I hope to have a good debate on the PR. We made a decision a while back to not use the main release branch of LinkML, as a result we forked it at a point in time and then loaded LinkML manually into this repo.

Then we stopped maintaining our fork.

The Fork [1] is 2070 commits behind main, has not been udpated in 3 years, and is full of vulnerabilities.
The vulns are not actually likely to be a source of risk given how we use this to build a static site, but the risk is not zero.
I propose we revert this model, go back to usual supply chain approaches and archive the fork.
Anyone who can dig out the reasons we did this way back when would be great to review those before we merge

[1]: https://github.com/govuk-one-login/linkml


huwd:
NB, something strange happened when I was trying to push this.
I pushed from main, something I don't usually worry about due to branch protection - seemed to have got through.

Will trigger a review of rulesets tomorrow and flag myself to folks who should know.
but I believe it's all fixed now.

Had to do some shady stuff to fix it though

huwd:
Ok so this still fails,
Noting @galund's commit here:
https://github.com/govuk-one-login/data-vocab/commit/4a9f938a7a60f7ae382806873631c8163f3e7b0f

Now to track down the bug

Joe-Edwards-GDS:
The differences can be seen at https://github.com/linkml/linkml/compare/main...govuk-one-login:linkml:data-vocab

Looks like three things
- Some work by Chris Midwood - ~presumably the original source of the fork, but I don't know exactly what it's doing~
    - original fork was work by George L, which is no longer used
- https://github.com/govuk-one-login/linkml/pull/1 - some work by me, required to generate Java classes that follow a reasonable inheritance structure (e.g. so `IdentityCheckCredential` extends `VerifiableCredential` etc.)
- https://github.com/govuk-one-login/linkml/pull/2 - some work by Chris C to format URIs correctly in json schema

To revert back to main, we'll either need to submit these as proper contributions to linkml (probably requiring feature flags etc., because they are breaking!) or come up with some other workaround for the features they cover - e.g. maintaining our own implementation of `jsonschemagen.py` and `markdownschemagen.py`, or some kind of post-processing hack.

Otherwise we'll have to keep this fork up to date to stay on top of vulnerabilities, which sounds very tedious unless there's an easy way to automate.

Joe-Edwards-GDS:
Superseded by #147

PR #147, opened 2025-11-03
Customise linkml

Joe-Edwards-GDS:
Update of https://github.com/govuk-one-login/data-vocab/pull/145/files to get JSON schema and markdown generation working properly.

- Added `jsonschemagen_wrapper` is added to wrap the base `jsonschemagen` to include hooks to replicate forked behaviour.
- Switched to `docgen`, replacing the deprecated `markdowngen`

While this PR aims to maintain compatibility of the schemas, there are a few differences:

## JSON schema

- The schema files use the updated `https://json-schema.org/draft/2019-09/schema` version of the meta-schema.
- The default behaviour also adds `null` as a valid value for any non-required properties, but I have reverted this so it can be added in a separate PR for visibility.

## Java / TS types

- The subtypes include all fields from the supertype (I couldn't easily replicate the old behaviour to exclude them). The java generator seems to ignore them if they match, but TS includes the duplicate (but this shouldn't actually make any difference to compilation, given the nature of TS).

## Markdown/HTML

`docgen` uses somewhat different templates to `markdowngen`, but it looks largely like a superset of the information:
- Separation by type (class vs enum etc.) is included natively
- There are some extra sections, e.g. linkml source
- The diagram is an interactive mermaid diagram, rather than static image

Example (see https://vocab.account.gov.uk/v1/classes/IdentityCheckCredentialClass/ for comparison):
_Users_joe edwards_dev-workspace_joee_data-vocab_site_v2_classes_IdentityCheckCredentialClass_index html

### DIFF
--- a/code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomObjectRule.java
+++ b/code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomObjectRule.java
@@ -108,19 +108,22 @@ public class CustomObjectRule extends ObjectRule {
 
         HashMap<String, JType> foundOverrides = new HashMap<>();
 
-        toStream(node.get("properties").fieldNames()).forEach(fieldName -> {
-            var field = superClass.fields().get(fieldName);
-            if (field == null) {
-                return;
-            }
+        var properties = node.get("properties");
+        if (properties != null) {
+            toStream(properties.fieldNames()).forEach(fieldName -> {
+                var field = superClass.fields().get(fieldName);
+                if (field == null) {
+                    return;
+                }
 
-            String pathToProperty = "#" + schema.getId().getFragment() + "/properties/" + fieldName;
+                String pathToProperty = "#" + schema.getId().getFragment() + "/properties/" + fieldName;
 
-            Schema propertySchema = this.ruleFactory.getSchemaStore().create(schema.getParent().getParent(), pathToProperty, this.ruleFactory.getGenerationConfig().getRefFragmentPathDelimiters());
-            JType propertyType = this.ruleFactory.getSchemaRule().apply(fieldName, propertySchema.getContent(), parent, jClass, propertySchema);
+                Schema propertySchema = this.ruleFactory.getSchemaStore().create(schema.getParent().getParent(), pathToProperty, this.ruleFactory.getGenerationConfig().getRefFragmentPathDelimiters());
+                JType propertyType = this.ruleFactory.getSchemaRule().apply(fieldName, propertySchema.getContent(), parent, jClass, propertySchema);
 
-            foundOverrides.put(field.type().name(), propertyType);
-        });
+                foundOverrides.put(field.type().name(), propertyType);
+            });
+        }
 
         JDefinedClass narrowedObject = jClass;
         for(var typeParam : superType.typeParams()) {
--- a/linkml_wrapper/jsonschemagen_wrapper.py
+++ b/linkml_wrapper/jsonschemagen_wrapper.py
@@ -1,0 +1,110 @@
+import click
+
+from linkml.generators.jsonschemagen import JsonSchemaGenerator, json_schema_types
+from linkml.generators.common.lifecycle import TClass
+from linkml.utils.generator import shared_arguments
+
+from linkml_runtime import SchemaView
+from linkml_runtime.linkml_model.meta import ClassDefinition
+
+# Add a URI type
+json_schema_types["uri"] = ("string", "uri")
+
+class JsonSchemaGeneratorWrapper(JsonSchemaGenerator):
+    def __init__(self, yamlfile, nonstandard_extends, *args, **kwargs):
+        super().__init__(yamlfile, *args, **kwargs)
+        self.nonstandard_extends = nonstandard_extends
+        # By default this will include null as a valid value for all non-required properties
+        self.include_null = False
+
+    def after_generate_class(self, cls: TClass, sv: SchemaView) -> TClass:
+        # By setting the non-standard 'extends' property
+        # jsonschema2pojo can generate Java classes with sensible inheritance hierarchy
+        if self.nonstandard_extends and cls.source.is_a is not None:
+            cls.schema_['extends'] = {
+                '$ref': f'#/$defs/{cls.source.is_a}'
+            }
+        return cls
+
+@shared_arguments(JsonSchemaGenerator)
+@click.command(name="json-schema")
+@click.option(
+    "-i",
+    "--inline",
+    is_flag=True,
+    help="""
+Generate references to types rather than inlining them.
+Note that declaring a slot as inlined: true will always inline the class
+""",
+)
+@click.option(
+    "-t",
+    "--top-class",
+    help="""
+Top level class; slots of this class will become top level properties in the json-schema
+""",
+)
+@click.option(
+    "--not-closed/--closed",
+    default=True,
+    show_default=True,
+    help="""
+Set additionalProperties=False if closed otherwise true if not closed at the global level
+""",
+)
+@click.option(
+    "--include-range-class-descendants/--no-range-class-descendants",
+    default=False,
+    show_default=False,
+    help="""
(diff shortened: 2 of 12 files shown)


code-generators/java-types/buildSrc/src/main/java/uk/gov/di/codegen/CustomObjectRule.java
```diff
@@ -108,19 +108,22 @@ private void addDefaultGenericOverrides(JDefinedClass jClass, JsonNode node, Jso
 
         HashMap<String, JType> foundOverrides = new HashMap<>();
 
-        toStream(node.get("properties").fieldNames()).forEach(fieldName -> {
-            var field = superClass.fields().get(fieldName);
-            if (field == null) {
-                return;
-            }
+        var properties = node.get("properties");
+        if (properties != null) {
```
Joe-Edwards-GDS, line 112: The new schema generator omits `properties` if it's empty, so we need a null guard here

pyproject.toml
```diff
@@ -1,23 +1,29 @@
...
 [tool.poetry]
 name = "di-identity-vocab"
 version = "1.0.0"
 description = "GOV.UK identity vocabulary"
 authors = ["Your Name <you@example.com>"]
 license = "MIT Copyright (c) 2022 Crown Copyright (Government Digital Service)"
 package-mode = false
+packages = [
+  { include = "scripts" }
+]
```
Joe-Edwards-GDS: This emits a big warning every time I run the script, so this is not the 'correct' way of doing this... Very keen for someone familiar with poetry to tell me what to do instead...

linkml_wrapper/jsonschemagen_wrapper.py
```diff
@@ -0,0 +1,110 @@
...
+
+from linkml.generators.jsonschemagen import JsonSchemaGenerator, json_schema_types
+from linkml.generators.common.lifecycle import TClass
+from linkml.utils.generator import shared_arguments
+
+from linkml_runtime import SchemaView
+from linkml_runtime.linkml_model.meta import ClassDefinition
+
+# Add a URI type
+json_schema_types["uri"] = ("string", "uri")
```
Joe-Edwards-GDS, line 11: This replicates the change from https://github.com/govuk-one-login/linkml/pull/2

linkml_wrapper/jsonschemagen_wrapper.py
```diff
@@ -0,0 +1,110 @@
...
+
+class JsonSchemaGeneratorWrapper(JsonSchemaGenerator):
+    def __init__(self, yamlfile, nonstandard_extends, *args, **kwargs):
+        super().__init__(yamlfile, *args, **kwargs)
+        self.nonstandard_extends = nonstandard_extends
+        # By default this will include null as a valid value for all non-required properties
+        self.include_null = False
```
Joe-Edwards-GDS, line 18: Without this the schema generates all non-required types as (e.g.) `[string, null]`, whereas previously it would just be `string`.

The new behaviour may be technically more correct - in which case we could remove this. Mostly affects T&R I expect (cc @chrisclayson )

linkml_wrapper/jsonschemagen_wrapper.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def after_generate_class(self, cls: TClass, sv: SchemaView) -> TClass:
+        # By setting the non-standard 'extends' property
+        # jsonschema2pojo can generate Java classes with sensible inheritance hierarchy
+        if self.nonstandard_extends and cls.source.is_a is not None:
+            cls.schema_['extends'] = {
+                '$ref': f'#/$defs/{cls.source.is_a}'
+            }
```
Joe-Edwards-GDS, line 26: This replicates the bodge I added - where we set `extends: { $ref: <some parent type> }` specifically when generating Java types

linkml_wrapper/jsonschemagen_wrapper.py
```diff
@@ -0,0 +1,110 @@
...
+        return cls
+
+@shared_arguments(JsonSchemaGenerator)
```
Joe-Edwards-GDS, line 35: Everything below here is pretty much replicating the CLI from the original (with the extra nonstandard_extends flag). Not sure if there's a cleverer way to 'inherit' these annotations.

v1/linkml-schemas/evidence.yaml
```diff
@@ -67,6 +67,7 @@ classes:
       - document
       - mitigation
       - incompleteMitigation
+      - txn
     slot_usage:
       txn:
```
Joe-Edwards-GDS, line 72: My read of the linkml docs is that `slots` is used to define which slots (i.e. common properties) are present, and `slot_usage` allows you to refine that in some way - so you can't have `slot_usage` without already having a `slot`. In any case it breaks if you don't do this!

scripts/test
```diff
@@ -2,13 +2,13 @@
...
+ajv validate --spec=draft2019 -c ajv-formats -s $HERE/../v1/json-schemas/IdentityCheckCredentialJWT.json -d $HERE/../v1/examples/app-gdc-liveness-likeness-jwt.json
+ajv validate --spec=draft2019 -c ajv-formats -s $HERE/../v1/json-schemas/IdentityCheckCredentialJWT.json -d $HERE/../v1/examples/identity-check-jwt.json
+ajv validate --spec=draft2019 -c ajv-formats -s $HERE/../v1/json-schemas/SecurityCheckCredentialJWT.json -d $HERE/../v1/examples/security-check-jwt.json
+ajv validate --spec=draft2019 -c ajv-formats -s $HERE/../v1/json-schemas/IssuerAuthorizationRequest.json -d $HERE/../v1/examples/core-cri-authz-request.json
+ajv validate --spec=draft2019 -c ajv-formats -s $HERE/../v1/json-schemas/OpenIDConnectAuthenticationRequest.json -d $HERE/../v1/examples/rp-authz-request.json
+ajv validate --spec=draft2019 -c ajv-formats -s $HERE/../v1/json-schemas/DeathRegisteredJWT.json -d $HERE/../v1/examples/life-events/death-registered.json
+ajv validate --spec=draft2019 -c ajv-formats -s $HERE/../v1/json-schemas/DeathRegistrationUpdatedJWT.json -d $HERE/../v1/examples/life-events/death-registration-updated.json
```
Joe-Edwards-GDS, line 14: @chrisclayson another thing - the new version emits with the `https://json-schema.org/draft/2019-09/schema` version (as opposed to version 07). Do you think that'll cause any issues?

linkml_wrapper/jsonschemagen_wrapper.py
chrisclayson: This would certainly solve some of our issues ... I'd want someone more (data) architectural to give a thumbs up on allowing `null` values for non-required fields (rather than them being not included).

From an implementation perspective it make very little difference, but there might be reasons why its better to not have `null` values 🤷🏻‍♂️

scripts/test
chrisclayson: I'd imagine this'll be fine, but let me double check schema version support in our tooling.

linkml_wrapper/jsonschemagen_wrapper.py
hugh-emerson: I looked at updating the linkml version in the event-catalogue a while ago and I found include_null didn't work properly. Slots that have `multivalued: true` still ended up with `"type": [ "array", "null"]`

(and then I gave up)

Does that happen here too?

EDIT: actually nevermind, they maybe fixed that issue according to the release notes for 1.9.5

Joe-Edwards-GDS: I did a diff of the old/new schemas which is how I spotted the issue, but not sure about this specific case(!)

hugh-emerson: I've tested out 1.9.5 now and it seems good to me.

fewtrees-pa (commented):
@Joe-Edwards-GDS - Who do you need to approve this PR?

chrisclayson (approved):

PR #213, opened 2026-04-09
SPT-1909: Publish to NPM registry

chrisclayson:
## What

- Update to use Node 24 (latest NPM required to use NPM Trusted Publishing)
- Refactor publishing steps into a shared action
- Add steps to publish to NPMjs as well as GitHub Packages (the latter to removed at a later date).
- Introduce use of `npmjs-publishing` GitHub Environment to lock down publishing only from `main` (manually configured):
image

## Why

We wish to publish NPM rather than GHP to lower the barrier to adoption (GHP requires authentication and permissions setting up).


.github/actions/npm-publish/action.yaml
```diff
@@ -0,0 +1,63 @@
...
+        scope: ${{ inputs.auth-scope }}
+
+    - name: Publish npm package to GitHub Packages
+      if: inputs.github-packages-token != ''
+      working-directory: ${{ inputs.artifact-path }}
+      shell: bash
+      run: |
+        npm publish
+      env:
+        NODE_AUTH_TOKEN: ${{ inputs.github-packages-token }}
```
danielmason-gds: Nit picking, missing EOL at EOF 😄

danielmason-gds (dismissed):
One thing, thats not quite a nit pick but is not blocking: it might be worth paramaterising the node version as you use it a few times and I'm sure this time next year we'll want to make it 26 😄

chrisclayson:
> One thing, thats not quite a nit pick but is not blocking: it might be worth paramaterising the node version as you use it a few times and I'm sure this time next year we'll want to make it 26 😄

Good point, well made ... just remembered I also wanted to SHA pin the action versions too 🤦🏻

.github/actions/npm-publish/action.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+    - name: Download TypeScript types
+      uses: actions/download-artifact@v8
+      with:
+        name: ${{ inputs.artifact-name }}
+        path: ${{ inputs.artifact-path }}
+
+    - uses: actions/setup-node@v6
+      with:
+        node-version: '24'
```
chris-dco: It may be better to use a node version file, then you don't need to duplicate it.

.github/actions/npm-publish/action.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - name: Publish npm package to NPMJS
+      working-directory: ${{ inputs.artifact-path }}
+      shell: bash
+      run: |
+        npm publish --provenance
+
+    - uses: actions/setup-node@v6
+      if: inputs.github-packages-token != ''
+      with:
+        node-version: '24'
```
chris-dco: As above. It may be better to use a node version file.

.github/workflows/build-test.yml
```diff
@@ -26,7 +26,7 @@ jobs:
           submodules: recursive
       - uses: actions/setup-node@v6
         with:
-          node-version: 22
+          node-version: 24
```
chris-dco: As above. It may be better to use a node version file.

.github/actions/npm-publish/action.yaml
chrisclayson: Fixed

.github/actions/npm-publish/action.yaml
chrisclayson: Good point ... change pushed.

danielmason-gds (approved):

chris-dco (approved):

PR #279, opened 2026-08-11
BAU: Fix security alerts

umaar:
## What

Bumps deps to fix some security alerts


chris-dco (approved):
Before and after merging this could you make sure that the pages are working properly such as https://vocab.account.gov.uk/v1/classes/AddressAssertionClass/. In some cases tables and diagrams have appeared as JSON data after updating some dependencies.

As a team we really need to show this repo some love. Thanks for looking at it.

